PR #28, opened 2022-12-13
revert changes for MWPW-120007

msagolj:
Minimal Browser versions for Milo to run is higher then minimal version for DC Widget, so no check required


msagolj:
As discussed with @Blainegunn we will  keep our version of browser check and update it to check fo the minimal browser version required for milo as this is higher then the DC widget requirments

PR #54, opened 2023-01-11
Mwpw 120007 Legacy Browser redirect

msagolj:
Checks if Browser is:
- Safari >= 14
- Edge >= 86

otherwise redirect to https://acrobat.adobe.com/home/index-browser-eol.html

Before : https://main--acrobat--adobecom.hlx.page/acrobat/online/pdf-to-ppt
After : https://mwpw-120007--acrobat--adobecom.hlx.page/acrobat/online/pdf-to-ppt

Note for testing:
- Testing lower then Safari 13.1 (labeled as 13 in SauceLabs) or Edge  80 will not work, as milo itself will fail to load.


Blainegunn (commented):
lgtm!

PR #57, opened 2023-01-16
Alloy Review Event

msagolj:
Before: https://main--acrobat--adobecom.hlx.page/acrobat/online/pdf-to-ppt
After: https://mwpw-122815--acrobat--adobecom.hlx.page/acrobat/online/pdf-to-ppt

To test:
- load the  page
- click on one of the rating stars at the bottom
- the console tab should show ```Review ALLOY event triggered```
- in network tab there should be a request triggered ```/collect?...```
- click on the request and check the ```payload``` tab
=> there should be an entry unter ```/events/0/data/_adobe_corpnew/digitaldata/primaryEvent/eventname``` set to ```productRating```


Blainegunn (changes requested):
perfect! 
but there's a second part. Look at the [alloyFeedbackEvent](https://git.corp.adobe.com/wcms/doccloud/blob/develop/app/jcr_root/apps/doccloud/clientlibs/base/react/src/js/components/review/ReviewDOM.js) 
We'll want to send 2 events for this block

PR #68, opened 2023-01-24
conditionally pre-render snippet only for first time users, set flag …

peller:
…to hydrate on script rather than content


peller:
@Blainegunn still in draft, will need to update widget version when it's available, then we can try this out

PR #70, opened 2023-01-25
use DOMParser instead of innerHTML

peller:
The pre-rendered document we are fetching is actually an HTML document with both a Style tag and a div with the content.  In order for React hydration to work, the DOM needs to match what is ultimately rendered, which means just the div part needs to be added to the widgetContainer.  We can put the Style tag directly in the head.

@Blainegunn


peller:
lookin' good.  FCP=LCP.  hydrate is working.
Screen Shot 2023-01-29 at 1 35 33 PM

PR #78, opened 2023-01-27
MWPW-123573

Blainegunn:


Blainegunn:
@apganapa @msagolj for your awareness. I've renamed the `analytics` folder to `alloy` so that it's not blocked by ad blockers

PR #102, opened 2023-02-01
MWPW-123750 fix salmon background

msagolj:
See https://jira.corp.adobe.com/browse/MWPW-123750
Fixed salom background to always be large enough in desktop, tablet, mobile view so that the widget does not overlap the  text that comes afterwards:

Before: https://main--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt
After: https://mwpw-123750--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

### DIFF
--- a/acrobat/blocks/dc-converter-widget/dc-converter-widget.css
+++ b/acrobat/blocks/dc-converter-widget/dc-converter-widget.css
@@ -14,6 +14,7 @@ min-height: 560px;
   position: relative;
   overflow: hidden;
   left:0;
+  top: 30px;
   right:0;
 }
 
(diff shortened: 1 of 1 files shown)


acrobat/blocks/dc-converter-widget/dc-converter-widget.css
```diff
@@ -11,21 +12,29 @@
...
 }
 
 #VERB, .hide {
   display: none;
 }
 
+
 .dc-wrapper {
   overflow: hidden;
+  width: 75%;
```
peller, line 28: does this match the XD for width?

PR #110, opened 2023-02-02
MWPW-124903 add missing config for exiting locales

msagolj:
Added missing locale entries for the currently supported regions. Copied the list from [milo](https://github.com/adobecom/milo/blob/main/libs/scripts/scripts.js#L23-L105)


Blainegunn (approved):
Best part of using GitHub… I can review PRs on my phone while I eat breakfast!

PR #129, opened 2023-02-13
MWPW-125486 Set explicit color for Accordion button to black

msagolj:
Font in FAQ section colored in blue instead of black. The issue is seen on both mobile & iPad devices.
Looks like some weird apple specific behavior on their devices, see e.g. https://developer.apple.com/forums/thread/690529.
Setting the color explicity solves the issue:

IPad

ipad

IPhone

iphone

Before: https://main--dc--adobecom.hlx.page/drafts/msagolj/live-test-pdf-to-ppt
After: https://mwpw-125486--dc--adobecom.hlx.page/drafts/msagolj/live-test-pdf-to-ppt
NOTE:
To Test you have to use a real apple device or one from saucelabs: https://app.saucelabs.com/live/web-testing/device

### DIFF
--- a/acrobat/styles/styles.css
+++ b/acrobat/styles/styles.css
@@ -338,3 +338,7 @@ div.how-to ol li::before{
   top: 0;
   left: 4px;
 }
+
+.accordion dt button {
+  color: #000;
+}


Blainegunn (approved):

Blainegunn:
getting it done!!!

PR #145, opened 2023-03-01
Mwpw 125812 update

SonjaPopovic:
Updates version for generate-cache to  correct version (1.163.1 vs 1.165.0) - per @JFernandezAdobe request on the ticket.

Added mappings for dynamically generated redirect urls (since for some verbs they are not following pattern and VERB.split('-').join('') would not give correct result, for example: for verb crop-pages, redirect url should be https://www.adobe.com/go/acrobat-crop and not https://www.adobe.com/go/acrobat-croppages, as for verbs extract-pages,combine-pdf,protect-pdf

### DIFF
--- a/acrobat/blocks/dc-converter-widget/dc-converter-widget.js
+++ b/acrobat/blocks/dc-converter-widget/dc-converter-widget.js
@@ -2,7 +2,22 @@ import frictionless from '../../scripts/frictionless.js';
 import { redirectLegacyBrowsers } from '../../scripts/legacyBrowser.js';
 
 const pageLang = document.querySelector('html').lang;
-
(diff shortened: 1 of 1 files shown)


Blainegunn (approved):
amazing!

JFernandezAdobe:
Hi @SonjaPopovic, @Blainegunn,

The redirect for all the verbs are looking good except for
https://mwpw-125812-update--dc--adobecom.hlx.page/acrobat/online/compress-pdf

The console prints https://www.adobe.com/go/acrobat-compresspdf-dev as the redirect link but it should be
https://www.adobe.com/go/acrobat-compress-dev

Please check.

PR #146, opened 2023-03-01
MWPW-127295

Blainegunn:
Resolves : [MWPW-127295](https://jira.corp.adobe.com/browse/MWPW-127295)
QA - https://mwpw-127295--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt
QA + MILO Fast Track https://mwpw-127295--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt?milolibs=fast-track


acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -1,7 +1,7 @@
 import frictionless from '../../scripts/frictionless.js';
 import { redirectLegacyBrowsers } from '../../scripts/legacyBrowser.js';
 
-const pageLang = document.querySelector('html').lang;
+const pageLang = document.querySelector('html').lang || 'en-US';
```
peller, line 4: This is a workaround. The lang attribute is missing this early. We need another way to pass in the locale.

Blainegunn: Yes, that will happen before the ROW rollout

acrobat/scripts/scripts.js
```diff
@@ -206,4 +199,15 @@ const miloLibs = setLibs(LIBS);
       window.dispatchEvent(imsIsReady);
     }
   }, 1000);
+
+  loadScript('/acrobat/scripts/bowser.js');
```
peller, line 203: have we considered whether we could live without bowser? UA sniffing should be very limited at this point. Perhaps we could hardcode a couple of rules for IE, Edge?

peller (approved):

JFernandezAdobe:
Hi @Blainegunn, @amauch-adobe 

It looks like all our tests in hlx.page and hlx.live failed.
```
  66 failed
  1 flaky
  65 passed (49.0m)
```

Sample test URL:
https://mwpw-127295--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

There is a 403 error in Github VMs (outside Adobe VPN) when the page tries to load the converter widget from DEV:
https://dev.acrobat.adobe.com/dc-hosted/2.37.2_1.165.0/dc-app-launcher.js

This is not an issue in the `main` branch, which loads the PROD converter widget:
https://main--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

peller:
@JFernandezAdobe running from the branch, unlike main, you need to be inside the firewall or the widget won't work (because it runs from dev). That is expected.  Do the tests run outside the firewall?

JFernandezAdobe:
> Do the tests run outside the firewall?

Yes. The tests running in Github Actions can only access publicly-accessible URLs. Anyway, I ran the tests manually in my local machine while connected to VPN, and everything looks good.

amauch-adobe:
> 

@JFernandezAdobe, it looks like someone has added a new rule set that any hlx.page URL is only accessible using VPN now.  There is a slack conversation about it, but it also has blocked Akamai from accessing things and shouldn't have been implemented according to Perry Wagner.  I'll keep you in the loop with what comes of the issue and its resolution.

PR #148, opened 2023-03-02
MWPW-120009-get widget and generate cache versions from metadata 

SonjaPopovic:
The idea is to use metadata.xslx file in dc root folder, to add  two meta tag to all pages under dc, for next attributes:
1) dc-widget-version
2) dc-generate-cache-version
So we need to add one more row in metadata file, to match all pages under dc folder, and 2 new columns for mentioned attributes, with values of needed widget version and generate cache versions. If those are not specified, we will use falback versions that are hardcoded in codebase


SonjaPopovic:
Tested with sidekick:  https://mwpw-120009-update-dc-version-using-metadata--dc--adobecom.hlx.page/drafts/sonja/converter

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -20,21 +20,24 @@ const verbToRedirectLinkSuffix =  {
 };
 export default function init(element) {
   const widget = element;
-  let WIDGET_ENV = 'https://dev.acrobat.adobe.com/dc-hosted/2.37.2_1.165.0/dc-app-launcher.js';
+  let DC_WIDGET_VERSION_FALLBACK = '2.37.2_1.165.0';
+  let DC_GENERATE_CACHE_VERSION_FALLBACK = '1.165.0';
+  let DC_WIDGET_VERSION = document.querySelector('meta[name="dc-widget-version"]')?.getAttribute('content') || DC_WIDGET_VERSION_FALLBACK;
+  let DC_GENERATE_CACHE_VERSION = document.querySelector('meta[name="dc-generate-cache-version"]')?.getAttribute('content') || DC_GENERATE_CACHE_VERSION_FALLBACK;
```
peller, line 26: use `const` where possible

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -20,21 +20,24 @@ const verbToRedirectLinkSuffix =  {
 };
 export default function init(element) {
   const widget = element;
-  let WIDGET_ENV = 'https://dev.acrobat.adobe.com/dc-hosted/2.37.2_1.165.0/dc-app-launcher.js';
+  let DC_WIDGET_VERSION_FALLBACK = '2.37.2_1.165.0';
+  let DC_GENERATE_CACHE_VERSION_FALLBACK = '1.165.0';
+  let DC_WIDGET_VERSION = document.querySelector('meta[name="dc-widget-version"]')?.getAttribute('content') || DC_WIDGET_VERSION_FALLBACK;
```
peller, line 25: providing a fallback means we won't notice a setup error.  Should we fail instead or log the error?

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -20,21 +20,24 @@ const verbToRedirectLinkSuffix =  {
...
+  let WIDGET_ENV = `https://dev.acrobat.adobe.com/dc-hosted/${DC_WIDGET_VERSION}/dc-app-launcher.js`;
   let ENV = 'dev';
   let REDIRECT_URL = '';
   let DC_GENERATE_CACHE_URL = '';
-
   if (window.location.hostname === 'main--dc--adobecom.hlx.page'
     || window.location.hostname === 'main--dc--adobecom.hlx.live'
     || window.location.hostname === 'www.adobe.com') {
-    WIDGET_ENV = 'https://documentcloud.adobe.com/dc-hosted/2.37.2_1.165.0/dc-app-launcher.js';
+    WIDGET_ENV = `https://documentcloud.adobe.com/dc-hosted/${DC_WIDGET_VERSION}/dc-app-launcher.js`;
```
peller, line 34: this url is also available in the metadata, along with dev and prod variants

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -80,7 +83,7 @@ export default function init(element) {
   if (preRender) {
     (async () => {
       // TODO: Make dynamic
-      const response = await fetch(DC_GENERATE_CACHE_URL || `https://documentcloud.adobe.com/dc-generate-cache/dc-hosted-1.165.0/${VERB}-${pageLang.toLocaleLowerCase()}.html`);
+      const response = await fetch(DC_GENERATE_CACHE_URL || `https://documentcloud.adobe.com/dc-generate-cache/dc-hosted-${DC_GENERATE_CACHE_VERSION}/${VERB}-${pageLang.toLocaleLowerCase()}.html`);
```
peller, line 86: this url can be generated from metadata if you choose `content.dc-generate-cache.urls.{$ENV}`

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
SonjaPopovic: In which metadata file is that stored? i don't see it here https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BBC0B7771-604F-4E0F-8319-FA0D6AEEFCBD%7D&file=metadata.xlsx&action=default&mobileredirect=true

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
SonjaPopovic: We are using https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BBC0B7771-604F-4E0F-8319-FA0D6AEEFCBD%7D&file=metadata.xlsx&action=default&mobileredirect=true  to generate metadata, and that file can be changed by authors, so they can update to wanted version . 

We can fetch urls from https://acrobat.adobe.com/dc-hosted/private/info.json for dc  widget version that is set in excel metadata, but not sure is it really necessary, @Blainegunn ?

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
peller: This metadata, which I guess would need to be carried over to sharepoint somehow: https://acrobat.adobe.com/dc-hosted/private/info.json

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
peller: info.json should be the source of the versions as well.  this should be automated somehow.

Blainegunn: @peller baby steps

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
peller: ...but do we know if lana logging will be available this early?

Blainegunn: we can capture the error and send it when lana loads 
we already do this in other places @peller

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
SonjaPopovic: @Blainegunn @peller yes, that was part of our plan - second step would be to create rundeck job which will generate widget version and update excel sheet based on https://acrobat.adobe.com/dc-hosted/private/info.json

peller: Ok, then you may want to substitute the partial urls here instead of just the version number, but that can be a follow up

Blainegunn (approved):

PR #164, opened 2023-03-16
MWPW-127442 | Reading locale from path compared to DOM value

apganapa:
Old: https://mwpw-127442--dc--adobecom.hlx.page/de/acrobat/online/pdf-to-ppt
Existing:  https://stage--dc--adobecom.hlx.page/de/acrobat/online/pdf-to-ppt


Blainegunn (approved):
smooth

PR #165, opened 2023-03-16
Mwpw 127570 update fragment loading experience

zagi25:
Resolves: https://jira.corp.adobe.com/browse/MWPW-127570

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

After: https://mwpw-127570-update-fragmnet-loading-expirience--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

### DIFF
--- a/acrobat/blocks/personalization/personalization.js
+++ b/acrobat/blocks/personalization/personalization.js
@@ -11,6 +11,7 @@ let showAll = false;
 export default function init(element) {
   const container = element;
   const frags = Array.from(container.children);
+  const userTags = localStorage.getItem('user_tags');
 
   let secondConversion;
   let upsell;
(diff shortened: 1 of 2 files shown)


draganatrajkovic (approved):

SonjaPopovic (approved):

Blainegunn (approved):
looks great!

Blainegunn:
@zagi25 however this seems to hurt the page performance somewhat, I'm not sure we should actually do this. 
The question is do we value the UX more than the page performance or the other way around

zagi25:
@Blainegunn if you are talking about delay in fragments loading I was also concerned about that, but on production  it looks that everything loads much faster and that delay that happens when fragments are switched (from default to upsell) is about 1sec, and is caused by setInterval in scripts.js that I changed in this PR. Also on published stage things load faster, can we somehow test this on published stage to see if delay is acceptable ?

PR #171, opened 2023-03-21
MWPW-128493

Blainegunn:
Resolves : [MWPW-128493](https://jira.corp.adobe.com/browse/MWPW-128493)
Small refactor


acrobat/blocks/dc-converter-widget/localeMap.js
```diff
@@ -0,0 +1,37 @@
+const localeMap = {
+  'ca_fr': 'fr-FR',
```
seanchoi-dev, line 2: Interestingly, this map isn't same as the locales in config - https://github.com/adobecom/dc/blob/stage/acrobat/scripts/scripts.js#L28
(If it were same, we could get it via getConfig() from milolibs)

I'd still like to suggest to have this map in `/scripts/utils.js`  in case we reuse somewhere else with different name than `localeMap` to distinguish the difference between the locales from config.

If you think this map will be used only for `dc-converter-widget` block, don't worry about this comment.

seanchoi-dev (approved):
Looks good, See my comment below.

PR #173, opened 2023-03-22
MWPW-125811 Set Foundation for Unit Test

TsayAdobe:
- Copy and follow the web test runner configuration from https://github.com/adobecom/milo
- Add tests for dcLana.js and dc-converter-widget.js

### DIFF
--- a/test/blocks/dc-converter-widget/dc-converter-widget.test.js
+++ b/test/blocks/dc-converter-widget/dc-converter-widget.test.js
@@ -1,0 +1,34 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+import sinon from 'sinon';
+import { waitForElement, delay } from '../../helpers/waitfor.js';
+
+document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+const { default: init } = await import(
+  '../../../acrobat/blocks/dc-converter-widget/dc-converter-widget'
+);
+
+describe('dc-converter-widget block', () => {
+  before(() => {
+    const block = document.body.querySelector('.dc-converter-widget');
+    init(block);
+  });
+
+  it('shoud creates a DC converter widget block', async function () {
+    this.timeout(5000); // default 2000ms
+    const button = await waitForElement('#lifecycle-nativebutton');
+    expect(document.querySelector('#lifecycle-drop-zone')).to.be.exist;
+    expect(document.querySelector('#lifecycle-nativebutton')).to.be.exist;
+  });
+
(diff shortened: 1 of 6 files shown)


test/blocks/dc-converter-widget/dc-converter-widget.test.js
```diff
@@ -0,0 +1,34 @@
...
+);
+
+describe('dc-converter-widget block', () => {
+  before(() => {
+    const block = document.body.querySelector('.dc-converter-widget');
+    init(block);
+  });
+
+  it('shoud creates a DC converter widget block', async function () {
+    this.timeout(5000); // default 2000ms
```
seanchoi-dev, line 18: It seems we have error on timeout
image

I think we may want to use delay() instead.

test/scripts/dcLana.test.js
```diff
@@ -0,0 +1,71 @@
+import lanaLogging from '../../acrobat/scripts/dcLana';
+import * as sinon from 'sinon';
+import { expect } from '@esm-bundle/chai';
+
+describe('Test dcLana script', async () => {
+  let clock;
+
+  before(() => {
```
seanchoi-dev, line 8: What do you think to add console.log in case something goes wrong?
Something like this?
``` 
before(() => {
  sinon.spy(console, 'log');
});

after(() => {
  console.log.restore();
});

seanchoi-dev:
Good work, @TsayAdobe . I added few comments to consider.

test/blocks/dc-converter-widget/dc-converter-widget.test.js
TsayAdobe: The block is depending on the dev widget inside Adobe network.

test/scripts/dcLana.test.js
TsayAdobe: Is it the standard way in Milo? I can add that.

test/blocks/dc-converter-widget/dc-converter-widget.test.js
TsayAdobe: I added skip for now. Need to test code coverage upload to codecov.io.

test/scripts/dcLana.test.js
TsayAdobe: dcLana.js doesn't call `console.log()`. There is no point to spy it.

seanchoi-dev (approved):
The 25.48% coverage is pretty low but it will be good starting point.

zagi25 (approved):

Blainegunn (approved):

PR #190, opened 2023-04-03
Mwpw 122642

SonjaPopovic:
Pricing card component, with options and content ( prices, ctas, and disclaimers ) that will be toggled based on selected option.Prices and ctas can be created by ost tool and used as inline prices/ctas. 
Quantity selector is in process of development for milo, so once it is done there should be follow up ticket to make sure it works with pricing card component as buttons and prices are working.

Preview:
https://mwpw-122642--dc--adobecom.hlx.page/drafts/sonja/document

Document: 

https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7Bb6290682-8cc0-422e-8ef5-c43500339b63%7D&action=editnew 

see readme file for more details about component


Blainegunn:
first off looks amazing!
A few thoughts.
1. Maybe we can move all the pricing info into separate fragments. For example for a fragment for price, one for disclaimer and another for (footer) cta. This way the can be managed independently at a high level!
2. Then we could have another fragment that points to the 3 smaller fragments making a commerce experience. 
3. Lastly we could update the `pricing-card` block to have a 3rd column that references the commerce experience fragment. Ideally from here it could link the hide/show with the radio buttons 

Here's rough sketch of what I mean 
Untitled-1

SonjaPopovic:
Tnx @Blainegunn . 
Than maybe in next steps we can include qty selector, buttons and prices in pricing-card component itself,and then users just setup what they need to be showed for pricing card. 

Currently i don't knew how prices, buttons and qty selectors should be connected with commerce / m@s and what are plans for m@s  and milo, so preparing pricing card to work with commerce experience fragment would be good first step.

SonjaPopovic:
Closing since it is moved on new branch

PR #192, opened 2023-04-04
Minimize waterfall prior to fetching the widget

peller:
inline small modules and defer modules which aren't yet needed, such as review and browser redirect

@Blainegunn


TsayAdobe:
The script has multiple syntax errors and the browser won't load it.

peller:
@TsayAdobe where do I find those syntax errors?

TsayAdobe:
* langLocaleMap is undefined
* Callback for IMS:Ready should be async since there is await inside.
* Callback for Bowser:Ready should be async since there is await inside.

Try dc-converter-widget.test.js (remove ".skip") to see if the widget can be loaded.

peller:
see #201

PR #196, opened 2023-04-04
MWPW-129213

Blainegunn:
Resolves : [MWPW-129213](https://jira.corp.adobe.com/browse/MWPW-129213)

### DIFF
--- a/acrobat/scripts/dcLana.js
+++ b/acrobat/scripts/dcLana.js
@@ -1,4 +1,5 @@
 export default function lanaLogging() {
+  const fricPage = window.location.pathname.indexOf('online/') > 0;
   const lanaOptions = {
     sampleRate: 1,
   };
@@ -11,7 +12,7 @@ export default function lanaLogging() {
   window.addEventListener('DC_Hosted:Ready', () => {
     const LANA_DC_WIDGET = document.querySelectorAll('[class*=ErrorDisplay]');
     LANA_DC_WIDGET.forEach((lanaDcw) => {
-      if (lanaDcw.querySelector('[class*=ErrorDisplay]')) {
+      if (fricPage && lanaDcw.querySelector('[class*=ErrorDisplay]')) {
         lanaOptions.tags = 'Cat=DxDC_Frictionless,origin=milo';
         window.lana.log('DC Widget Failed ¶ Reason: Error', lanaOptions);
       }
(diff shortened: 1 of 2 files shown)


acrobat/scripts/dcLana.js
```diff
@@ -1,4 +1,5 @@
 export default function lanaLogging() {
+  const fricPage = window.location.pathname.indexOf('online/') > 0;
```
seanchoi-dev, line 2: indexOf is kinda old function.
I'd use includes() instead.
```suggestion
  if (window.location.pathname.includes('online/')) {
```
But really, I don't think we want this since it will be `true` for all any other URLs potentially includes `online`. This may result in unwanted result.

Can we actually move this into `frictionless.js` so it gets run only for frictionless?
Like move 
```
const { default: lanaLogging } = await import('./dcLana.js');
lanaLogging();
```
from scripts.js to frictionless.js?

acrobat/scripts/scripts.js
```diff
@@ -184,8 +184,10 @@ function decoratePromotion() {
   const { default: lanaLogging } = await import('./dcLana.js');
 
   // Setup CSP
-  const { default: ContentSecurityPolicy } = await import('./contentSecurityPolicy/csp.js');
-  ContentSecurityPolicy();
+  if (window.location.pathname.indexOf('online') > 0) {
+    const { default: ContentSecurityPolicy } = await import('./contentSecurityPolicy/csp.js');
+    ContentSecurityPolicy();
+  }
```
seanchoi-dev, line 190: Same as above, can we just move this code into `frictionless.js`?

seanchoi-dev (changes requested):

acrobat/scripts/scripts.js
Blainegunn: no, this needs to be in `scripts.js` for expedience

fullcolorcoder:
+1 on Sean's comments. Otherwise looks good to me

acrobat/scripts/dcLana.js
Blainegunn: it's `online/` not `online`. Everything in the directory has the widget.
I want all the LANA code together

acrobat/scripts/dcLana.js
amauch-adobe, line 1: I'm probably missing something, so feel free to steer me in the right direction if I am.  Is there a reason we want to turn off LANA for the FedPub pages in DC?  For site reliability, we wanted to add Lana logging to more things so the if a block or element on a page stops working exceptions and other messages could be logged to SPLUNK using LANA.  I worry removing it, even for FedPub pages that don't have much but links, images, and text could still be a detriment.  What are your thoughts?

amauch-adobe (commented):
@Blainegunn see my comment on the LANA Logging removal.

acrobat/scripts/dcLana.js
Blainegunn: LANA is still on FedPub pages just not the frictionless. We load it for the entire site in `scripts.js` for general errors.

amauch-adobe (approved):
Approved.

acrobat/scripts/scripts.js
```diff
@@ -184,8 +184,10 @@ function decoratePromotion() {
   const { default: lanaLogging } = await import('./dcLana.js');
 
   // Setup CSP
-  const { default: ContentSecurityPolicy } = await import('./contentSecurityPolicy/csp.js');
-  ContentSecurityPolicy();
+  if (window.location.pathname.indexOf('online') > 0) {
```
seanchoi-dev, line 187:
```suggestion
  if (window.location.pathname.includes('online/') {
```
Should do.. but again, just be aware that will be true any URL includes `online/`. 
I would be happier if we make this working with metadata instead. 

Something like..
```
const frictionlessMeta = getMetadata('frictionless');
  if (frictionlessMeta === 'on') {
```

seanchoi-dev (approved):
I approve this but please consider my comment below.

acrobat/scripts/scripts.js
seanchoi-dev: At least that's how milo core does when URL based condition is required...

Blainegunn: i hear you! let's do that then!

PR #198, opened 2023-04-05
Improved handling promotion block from metadata (for FedPub)

seanchoi-dev:
Resolve: [MWPW-129216](https://jira.corp.adobe.com/browse/MWPW-129216)

Before: https://stage--dc--adobecom.hlx.page/acrobat/hub/how-to/how-to-convert-pdf-to-image
After: https://delay-promotion--dc--adobecom.hlx.page/acrobat/hub/how-to/how-to-convert-pdf-to-image

This will load promotion block that being added when metadata exists after loadArea().
Note that loadArea() includes loadPostLCP which means **_this promotion block will be loaded after loadPostLCP()_**.
If you see refresh and go page bottom quickly you will find the promotion block get loaded slowly after everything else.

@Blainegunn  I don't know the background context why we want to load this after everything else, but this will do what is being asked.

### DIFF
--- a/acrobat/blocks/promotion/promotion.js
+++ b/acrobat/blocks/promotion/promotion.js
@@ -1,5 +1,3 @@
-import { getLibs } from '../../scripts/utils.js';
-
 export default async function init(el) {
   const locale = 'en-US';
   const promotionName = el.getAttribute('data-promotion');
--- a/acrobat/blocks/promotion/promotion.js
+++ b/acrobat/blocks/promotion/promotion.js
@@ -1,3 +1,5 @@
+import { getLibs } from '../../scripts/utils.js';
+
 export default async function init(el) {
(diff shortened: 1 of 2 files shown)


Blainegunn (approved):

fullcolorcoder (approved):

TsayAdobe:
Before: the block is init'ed by Milo, `promotion.css` is loaded.
After: the block is init'ed by scripts.js. `promotion.css` is not loaded.

seanchoi-dev:
> Before: the block is init'ed by Milo, `promotion.css` is loaded. After: the block is init'ed by scripts.js. `promotion.css` is not loaded.

Not loading the css file is intended since it is 0% we will need a custom CSS for promotion until we revisit this.

PR #199, opened 2023-04-05
Improving LCP first load time.

seanchoi-dev:
Moving the dcLana import right before the function get called so loadArea() doesn't have to wait the file get loaded.

before: https://stage--dc--adobecom.hlx.live/acrobat/online/pdf-to-ppt
after: https://lana--dc--adobecom.hlx.live/acrobat/online/pdf-to-ppt


Blainegunn (approved):

peller:
Thank you, @seanchoi-dev !

PR #205, opened 2023-04-11
MWPW-129451 reduce waterfall after prerendered widget

peller:
Minimize the time between loading the pre-rendered HTML and calling `loadArea` which makes it visible.

BEFORE 

Screen Shot 2023-04-10 at 10 47 41 PM

AFTER

Screen Shot 2023-04-10 at 10 24 18 PM


seanchoi-dev (approved):
This helps the waterfall, but this is likely disable the getLibs() from utils.js which may become an issue in future.
I am approving this because I can't think of a better idea for now.
Possible idea is move `export const [setLibs, getLibs]` function into `scripts.js` instead having setLibs() alone..

peller:
Thanks @seanchoi-dev  I cut it because it was not being used.  A good set of lint rules would probably complain if I had left it there :) we can add it back if needed.

peller:
It doesn't take too much imagination how to eliminate the remainder of the waterfall. We should look at using a tool like webpack or rollup to package up the code we need.

seanchoi-dev:
> Thanks @seanchoi-dev I cut it because it was not being used. A good set of lint rules would probably complain if I had left it there :) we can add it back if needed.

@peller The function `export const [setLibs, getLibs]` is given by milo-college which is standard for all milo consumer site including this site. It is used for getting Milo's library. I don't currently see any getLibs() used in DC, yet but it may need it soon since Milo has a lot of useful functions.

Example of usage: https://github.com/adobecom/milo-college/blob/main/scripts/utils.js#L40-L42

Blainegunn:
@peller resolved your conflicts

Blainegunn (approved):

PR #207, opened 2023-04-11
MWPW-129156 Fix verb footer style

TsayAdobe:
Resolves: https://jira.corp.adobe.com/browse/MWPW-129156

Before: https://www.adobe.com/acrobat/online/pdf-to-ppt

After: https://mwpw-129156--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

### DIFF
--- a/acrobat/styles/styles.css
+++ b/acrobat/styles/styles.css
@@ -106,35 +106,6 @@
   background: #fbfbfb;
 }
 
-.accordion-container.verb-subfooter-moblie .accordion-trigger {
-  color: #747474;
-  -ms-flex-align: center;
-  -ms-flex-pack: start;
-  align-items: center;
-  -webkit-appearance: none;
-  -moz-appearance: none;
-  appearance: none;
-  background-color: inherit;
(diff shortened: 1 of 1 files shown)


seanchoi-dev:
I am looking how Milo does and wondering if we should if from Milo side.

seanchoi-dev (changes requested):
Looks like this doesn't make the result as same as what consonant suggests.
image

I would actually remove this css block (.accordion-container.verb-subfooter-moblie .accordion-trigger { }
) since Milo already provides the consonant style as default.

TsayAdobe:
@seanchoi-dev My Chrome and Safari don't render the section like your screenshot.

seanchoi-dev:
> @seanchoi-dev My Chrome and Safari don't render the section like your screenshot. 

I am not sure which screenshot you're referring to... but my last screenshot is the result when I removed all css inside `.accordion-container.verb-subfooter-moblie .accordion-trigger { }`

Blainegunn (approved):
looks great @TsayAdobe

PR #219, opened 2023-04-19
site updates to move folder

Blainegunn:


helix-query.yaml
```diff
@@ -59,7 +59,7 @@ indices:
       - '**/drafts/**'
       - '**/fragments/**'
       - '**/localization/**'
-    target: /dc-shared/assets/sitemaps/query-index-de.xlsx
+    target: de/dc-shared/assets/sitemaps/query-index.xlsx
```
hparra, line 62: missing root `/`

hparra (changes requested):
See typo and please put a description in PR and some hlx.live link in description (just so you get that check to pass)

Addresses: https://jira.corp.adobe.com/browse/MWPW-129768

Test URL:

- (anything -- we'd put link to json but that hasn't been generated yet!)

helix-query.yaml
Blainegunn: Good catch

PR #222, opened 2023-04-20
MWPW-128882: Add og:locale meta tag

zagi25:
Resolves: [MWPW-128882](https://jira.corp.adobe.com/browse/MWPW-128882)

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt

After: https://mwpw-128882-user-ratings--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt


zagi25:
@Blainegunn I returned review part of frictionless.js to previous state because I made a mistake. I didn't know that 4th and 5th star are supposed to act differently from others. Sorry about that, I will make sure that in the future I ask more questions if something is unclear to me.

acrobat/scripts/frictionless.js
```diff
@@ -12,16 +12,24 @@ export default function init(verb) {
...
           reviewFeedbackAlloy(verb, data.rating, data['rating-comments']);
         });
+        if (document.querySelectorAll('.tooltip').length > 0) {
+          document.querySelectorAll('.tooltip')[3].addEventListener('click', () => {
+            reviewFeedbackAlloy(verb, '4');
+          })
+          document.querySelectorAll('.tooltip')[4].addEventListener('click', () => {
+            reviewFeedbackAlloy(verb, '5');
+          })
+        }
```
seanchoi-dev, line 32: It would be better to call querySelectorAll() just one time under reviewBlock element that is predeclared to save cost and avoid getting a wrong element in case there are more elements with `tooltip` class.
```suggestion
        const reviewTooltip = reviewBlock.querySelectorAll('.tooltip');
        if (reviewTooltip.length > 0) {
          reviewTooltip[3].addEventListener('click', () => {
            reviewFeedbackAlloy(verb, '4');
          })
          reviewTooltip[4].addEventListener('click', () => {
            reviewFeedbackAlloy(verb, '5');
          })
        }
```

acrobat/scripts/scripts.js
```diff
@@ -19,6 +21,17 @@ function loadStyles(paths) {
...
 }
 
+function addLocale() {
+  const url = new URL(window.location.href);
+  const langFromPath = url.pathname.split('/')[1];
+  const pageLang = localeMap[langFromPath] || 'en-US';
+  const metaTag = document.createElement('meta');
+  metaTag.setAttribute('property', 'og:locale');
+  metaTag.setAttribute('content', pageLang);
+  const head = document.querySelector('head');
```
TsayAdobe, line 31: Isn't document.head faster if the code doesn't need to work for older IE (IE < 9)

acrobat/scripts/scripts.js
```diff
@@ -10,6 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
+import localeMap from "./maps/localeMap.js";
```
seanchoi-dev, line 13:
```suggestion
import localeMap from './maps/localeMap.js';
```
It should be single quotation for string.

acrobat/scripts/scripts.js
```diff
@@ -19,6 +21,17 @@ function loadStyles(paths) {
   });
 }
 
+function addLocale() {
+  const url = new URL(window.location.href);
+  const langFromPath = url.pathname.split('/')[1];
```
seanchoi-dev, line 26:
```suggestion
  const { pathname } = window.location;
  const langFromPath = pathname.split('/')[1];
```
This is simpler way to get pathname from location.

TsayAdobe, line 26: We don't need the rest of splits, why not put a limit?
```
url.pathname.split('/', 2)[1];
```

TsayAdobe (approved):

seanchoi-dev (changes requested):
I left few suggestions.
But do we really need to use ./maps/localeMap.js instead locales that defined at scripts.js?
I see its `ietf` defined differently from `localeMap.js`, but I don't know the background story why we needed localeMap.js and if we want to use it for `og:locale`.

If we can use locales from scripts.js, than we could also use `getLocale()` from milo utils.js which is free.

zagi25:
@TsayAdobe @seanchoi-dev @Blainegunn PR is updated. Thanks for the help. I used getLocale() from milo, like Sean suggested, with locales from our script.js because I checked couple of prod pages with locale, and their og:locale corresponds to values in locales object.

Blainegunn (approved):

PR #227, opened 2023-04-26
Contentroot update

seanchoi-dev:
Preventing double //
This PR will merge directly into main since this is kinda urgent.


TsayAdobe (approved):
We may want to fix the error-prone code in Milo.
```
      config.locale.contentRoot = `${origin}${config.locale.prefix}${config.contentRoot ?? ''}`;
```

PR #233, opened 2023-05-04
Use franklin feature to add `x-default` hreflang

hparra:
Still unsure if `us` entry needs `alternate` key or not. Need to read source more carefully: https://github.com/adobe/helix-admin/blob/main/src/sitemap/builder.js

Addresses: [MWPW-130622](https://jira.corp.adobe.com/browse/MWPW-130622). May may not fix duplicates but at least it will add missing `x-default`

URL for testing:

- https://hparra-patch-2--dc--adobecom.hlx.page/

Post-merge:

- Run `curl -vX POST https://admin.hlx.page/sitemap/adobecom/dc/main/dc-shared/assets/sitemap.xml`

### DIFF
--- a/helix-sitemap.yaml
+++ b/helix-sitemap.yaml
@@ -4,13 +4,8 @@ sitemaps:
   website:
     origin: https://www.adobe.com
     lastmod: YYYY-MM-DD
+    default: en-US
     languages:
-      default:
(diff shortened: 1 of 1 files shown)


TsayAdobe (approved):
Looks good

fullcolorcoder (approved):

PR #253, opened 2023-06-03
DCXY-17389 Synchronize hydration with creation of dropzone

peller:
Resolves former MWPW [DCXY-17389](https://jira.corp.adobe.com/browse/DCXY-17389)
Prevent the case where the widget gets to the hydration step before the pre-rendered dropzone is inserted in the DOM.


acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -192,7 +194,7 @@ export default function init(element) {
   dcScript.dataset.load_imslib = 'false';
   dcScript.dataset.enable_unload_prompt = 'true';
   if (preRenderDropZone) {
-    dcScript.dataset.pre_rendered = 'true';
+    dcScript.dataset.pre_rendered = 'true'; // TODO: remove this line
```
Blainegunn, line 197: why?
or when?

peller: until we update the widget, we'll need this bit to do SSR with the old widget.  Going forward, it won't be needed anymore.  The widget will check for the bit on the CID element.

PR #258, opened 2023-06-11
MWPW-132502 remove z-index from CID, move .section behind to z-index -1

peller:
It would resolve this bug not to set the widget background z-index to 1, and it would seem to be far safer to avoid z-index conflicts with the react-spectrum library, in general.

Would moving all .section to z-index: -1 create any problems?


peller:
doesn't work. closing

PR #259, opened 2023-06-12
mwpw-132502: Added isolation property to main

zagi25:
Please note that the widget version is different on stage branch compared to dev branches.

Resolves: [132502](https://jira.corp.adobe.com/browse/MWPW-132502)
Before: https://stage--dc--adobecom.hlx.page/acrobat/online/pdf-to-jpg
After: https://mwpw-132502-widget-dropdonw-menu--dc--adobecom.hlx.page/acrobat/online/pdf-to-jpg

### DIFF
--- a/acrobat/blocks/personalization/personalization.css
+++ b/acrobat/blocks/personalization/personalization.css
@@ -32,3 +32,7 @@
   max-height: 0;
   overflow: hidden;
 }
+
+main:has(#dc-converter-widget) {
+  isolation: isolate;
+}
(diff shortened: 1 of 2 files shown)


Blainegunn:
@zagi25 your fix is spot but it won't work on firefox 
https://developer.mozilla.org/en-US/docs/Web/CSS/:has
I'm going to update the PR and add the CSS rules to the personalization.css (since it's only used on frictionless pages) and merge. thank you

Blainegunn (approved):

PR #261, opened 2023-06-13
Implemented fix

saragajic11:
The gnav sign-in option comes from dc-shared/gnav.docx has a specified link that exists for US but does not exist for other locations. A regular click on the option redirects as expected, but opening the link in a new tab gives a 404. This fix sets the href attribute value to "#" which will open the same page in a new tab, without requiring authors to update the gnav links for all locales.
Also, this PR resolves issue with section menu which currently redirects to wrong page on new tab, while it should redirect to same page.
Resolves: [MWPW-132208](https://jira.corp.adobe.com/browse/MWPW-132208) and [MWPW-132728](https://jira.corp.adobe.com/browse/MWPW-132728)
Before: https://stage--dc--adobecom.hlx.page/de/acrobat/online/jpg-to-pdf
After: https://mwpw-132208-fix-gnav-signin-link--dc--adobecom.hlx.page/de/acrobat/online/jpg-to-pdf

### DIFF
--- a/acrobat/scripts/scripts.js
+++ b/acrobat/scripts/scripts.js
@@ -208,7 +208,6 @@ const CONFIG = {
   addLocale(ietf);
   setConfig({ ...CONFIG, miloLibs });
   loadLana({ clientId: 'dxdc' });
-  await loadArea();
 
   // Setup href of sign in
   const singInReady = setInterval(() => {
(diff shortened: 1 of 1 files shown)


Blainegunn:
@saragajic11 if this a real issue for us, it is for all MILO consumers and we should fix it in their project.
Alternatively we can ask GWP to no localize the signin links so they redirect if they are opened in a new window/tab

PR #267, opened 2023-06-20
MWPW-129281 - CLS Regression

Blainegunn:
Resolves : [MWPW-129281](https://jira.corp.adobe.com/browse/MWPW-129281)

### DIFF
--- a/acrobat/blocks/eventwrapper/eventwrapper.js
+++ b/acrobat/blocks/eventwrapper/eventwrapper.js
@@ -21,6 +21,7 @@ const TRY_ANOTHER = 'try-another-file-start';
 
 export default function init(element) {
   const wrapper = element;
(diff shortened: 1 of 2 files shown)


acrobat/blocks/eventwrapper/eventwrapper.js
```diff
@@ -92,6 +93,7 @@ export default function init(element) {
         break;
       case UPLOAD_START:
         setCurrentEvent('upload');
+        if (reviewBlock[0]) { reviewBlock[0].classList.add('fade'); };
```
peller, line 96: you might want to hide it immediately on upload (simultaneously or prior to shifting the content upward) then fade it back in after the upload completes?  It might even be possible to do this with only CSS rules

PR #268, opened 2023-06-20
MWPW-132671: Load Milo's locales for config

hparra:
Supports: [MWPW-132671](https://jira.corp.adobe.com/browse/MWPW-132671)

Requires: https://github.com/adobecom/milo/pull/880

**Test URLs:**

- Before: https://use-milo-locales--dc--adobecom.hlx.live/ar/acrobat/online/word-to-pdf
- After: https://use-milo-locales--dc--adobecom.hlx.live/ar/acrobat/online/word-to-pdf?milolibs=import-locales--milo--hparra

### DIFF
--- a/acrobat/scripts/scripts.js
+++ b/acrobat/scripts/scripts.js
@@ -99,10 +99,14 @@ const CONFIG = {
   loadStyles(paths);
 
   // Import base milo features and run them
-  const { locales } = await import(`${miloLibs}/scripts/scripts.js`);
-  const {
-    loadArea, loadDelayed, loadScript, setConfig, loadLana, getMetadata, getLocale
-  } = await import(`${miloLibs}/utils/utils.js`);
+  const [
+    { locales },
+    { loadArea, loadDelayed, loadScript, setConfig, loadLana, getMetadata, getLocale }
+  ] = await Promise.all([
+    import(`${miloLibs}/scripts/locales.js`),
+    import(`${miloLibs}/utils/utils.js`),
+  ]);
+
   const { ietf } = getLocale(locales);
   addLocale(ietf);
   setConfig({ ...CONFIG, locales, miloLibs });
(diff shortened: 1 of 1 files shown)


acrobat/scripts/scripts.js
```diff
@@ -201,12 +99,13 @@ const CONFIG = {
   loadStyles(paths);
 
   // Import base milo features and run them
+  const { locales } = await import(`${miloLibs}/scripts/scripts.js`);
```
peller, line 102: this will make our page load slower.

peller (changes requested):
let's discuss performance ramifications of this change

acrobat/scripts/scripts.js
hparra: Change is in the microseconds. Will be loaded from `/scripts/locales.js` when change is made in Milo. This is POC right now.

peller: everything is microseconds.  hundreds of microseconds on a constrained connection.  If this is merged, won't it go live?

acrobat/scripts/scripts.js
```diff
@@ -201,12 +99,17 @@ const CONFIG = {
   loadStyles(paths);
 
   // Import base milo features and run them
-  const {
-    loadArea, loadDelayed, loadScript, setConfig, loadLana, getMetadata, getLocale
-  } = await import(`${miloLibs}/utils/utils.js`);
+  const [
+    { locales },
+    { loadArea, loadDelayed, loadScript, setConfig, loadLana, getMetadata, getLocale }
+  ] = await Promise.all([
```
peller, line 105: Doing this in parallel is better, but everything that comes before loadArea is critical, so we had inlined this stuff on purpose (even though duplicating code is not the ideal approach)

What is the motivation for changing this?

@Blainegunn

peller: ...ideally we'd have a rollup/webpack builder process our imports and do this for us.

hparra:
@peller See https://github.com/adobecom/milo/pull/880 and the issue it links to for context. I'll recreate this PR against staging if/when that lands.

PR #272, opened 2023-06-28
MWPW-133310

Blainegunn:
Resolves : [MWPW-133310](https://jira.corp.adobe.com/browse/MWPW-133310)


acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -193,6 +193,7 @@ export default function init(element) {
 
   const widgetContainer = document.createElement('div');
   widgetContainer.id = 'CID';
+  widgetContainer.className = `wapper-${VERB}`;
```
peller, line 196: typo, but it matches

PR #273, opened 2023-06-29
MWPW-133205 Update CSP

TsayAdobe:


acrobat/scripts/contentSecurityPolicy/dev.js
```diff
@@ -10,6 +10,7 @@ const connectSrc = [
   '*.clarity.ms',
   '*.clicktale.net/',
   '*.mktoresp.com',
+  '*.demandbase.com',
```
peller, line 13: do we need the wildcard here?  for security purposes, it's better not to

TsayAdobe: Updated to `tag-logger.demandbase.com`

peller (approved):

PR #283, opened 2023-07-18
MWPW-132031

trixyfox:
*Placeholder credit-cards will swap out images for text strings

Resolves: [MWPW-132031](https://jira.corp.adobe.com/browse/MWPW-132031)

Test URLs:

Before: https://main--dc--adobecom.hlx.page/drafts/steadams/credit-card
After: https://placeholder-cc--dc--trixyfox.hlx.page/drafts/steadams/credit-card


acrobat/scripts/ccTranslations.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  pTags.forEach((tag) => {
+    const ccimg = document.createElement('img');
+    const ccName = tag.innerHTML.toLowerCase();
+    const imgExt = '.jpg';
+    const imgPrefix = 'credit-cards-';
+    const imageName = imgPrefix + ccName + imgExt;
+    const imagePath = `/acrobat/img/icons/${imageName}`;
+    if (checkIfImageExists(imagePath)) {
```
TsayAdobe, line 23: Do we have to check if image exists? It will take more network trips.

TsayAdobe (commented):
Tests are not done yet.

acrobat/scripts/scripts.js
```diff
@@ -200,9 +200,26 @@ const CONFIG = {
   if (STYLES) { paths.push(STYLES); }
   loadStyles(paths);
 
+  // Labeling the placeholders before they get processed by Milo
+  const placeholderLabel = (document) => {
```
Blainegunn, line 204: I like this

acrobat/scripts/scripts.js
```diff
@@ -200,9 +200,26 @@ const CONFIG = {
...
+      pTags.forEach((pTag) => {
+        if (pTag.innerHTML.match(ccregex)) {
+          const translationLabel = pTag.innerHTML;
+          const label = translationLabel.slice(2, -2);
+          pTag.setAttribute('data-local', label);
+        }
+      });
+    }
+  };
+  placeholderLabel(document.documentElement);
```
Blainegunn, line 218: might consider doing this the other way around to see if faster, but again I really like this approach

acrobat/scripts/scripts.js
```diff
@@ -250,3 +267,8 @@ const bowserReady = setInterval(() => {
     window.dispatchEvent(bowserIsReady);
   }
 }, 100);
+
+const { default: ccTranslations } = await import('./ccTranslations.js');
+document.addEventListener('milo:deferred', () => {
+  ccTranslations();
```
Blainegunn, line 273: you should pass the elements in this function so you don't have to search the DOM again

acrobat/scripts/ccTranslations.js
```diff
@@ -0,0 +1,33 @@
+export default function ccTranslations() {
```
Blainegunn, line 1: pass elements into this func to prevent having to find them again

acrobat/scripts/ccTranslations.js
```diff
@@ -0,0 +1,33 @@
...
+    const http = new XMLHttpRequest();
+    http.open('HEAD', url, false);
+    http.send();
+    if (http.status !== 404) {
+      return true;
+    }
+    return false;
+  }
+  function errorLogging() {
+    window.lana?.log('No image available for credit-card placeholder');
```
Blainegunn, line 13: very nice, also pass in the `imagePath` so we know what failed

Blainegunn: also pass in some logging details like we do everywhere else
```
lanaOptions.tags = 'Cat=DC_MILO_Commerce,origin=milo';
window.lana.log('No image available for credit-card placeholder ¶ img: imagePath', lanaOptions);
```

acrobat/scripts/ccTranslations.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  pTags.forEach((tag) => {
+    const ccimg = document.createElement('img');
```
Blainegunn, line 17: MILO has method for to create elements easier

Blainegunn (changes requested):
also there's no test so this can't be merged until it has a unit test

acrobat/scripts/ccTranslations.js
zagi25, line 23: Instead of checkIfImageExists function you can use error event
```
ccimg.onerror =  () => { ... }
```

acrobat/scripts/ccTranslations.js
seanchoi-dev: +1 it is called createTag()

acrobat/scripts/scripts.js
peller: Not sure if it matters, but does this need to happen before loadArea?

Blainegunn:
Archiving this until we need this feature, thanks 1000x @trixyfox for the excellent contribution

PR #293, opened 2023-07-28
Revert "Merge pull request #287 from adobecom/ajs-csp"

peller:
It was determined that the child-src:blob setting was not needed to support fill & sign L2. the worker-src:blob setting is sufficient because everything is running on the same domain.

This reverts commit 7a34bed8f9d2b6e1e32af344405792825d51fa2e, reversing changes made to 65eda6d2b928fc9ee8e82885fabc88f73e8161ca.


peller:
To support Safari < 15.4, it looks like we do need child-src after all. Closing this PR.

PR #294, opened 2023-07-31
Added padding for how-to block

saragajic11:
*Added padding to how-to block to fix overlapping on tablet devices + added additional padding for desktop devices to match previous design

Resolves: [MWPW-134277](https://jira.corp.adobe.com/browse/MWPW-134277)

Test URLs:

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/word-to-pdf
After: https://mwpw-134277-fix-overlap--dc--adobecom.hlx.page/acrobat/online/word-to-pdf

### DIFF
--- a/acrobat/styles/styles.css
+++ b/acrobat/styles/styles.css
@@ -205,6 +205,7 @@ div.how-to {
     "image" !important;
   }
   .how-to-heading {
+    padding-top: 44px;
     margin-right: 0;
   }
   .how-to-image.how-to-image-large {
@@ -393,11 +394,5 @@ div.how-to ol li::before{
 
   .how-to-heading {
     padding-top: 52px;
-  }
-}
-
-@media (max-width: 599px) {
-  .how-to-heading {
-    padding-top: 44px;
   }
 }
--- a/acrobat/styles/styles.css
+++ b/acrobat/styles/styles.css
@@ -188,6 +188,16 @@ div.how-to {
   margin-right: 28px;
 }
 
+@media (max-width: 1200px) {
+  .how-to > .foreground {
+    padding-top: 80px !important;
+  }
+
+  .how-to-heading {
+    padding-top: 52px;
+  }
+}
+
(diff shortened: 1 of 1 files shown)


acrobat/styles/styles.css
```diff
@@ -386,3 +387,9 @@ div.how-to ol li::before{
   margin: auto;
   padding: 20px;
 }
+
+@media (max-width: 1200px) {
+  .how-to-heading {
+    padding-top: 52px;
```
zagi25, line 393: This will add unnecessary padding on smaller screens. Widget changes height at screen width of 768px, so you can do this: 
```
@media (min-width: 768px) {
  .how-to-heading {
    padding-top: 52px;
  }
```

saragajic11: Hi @zagi25 thanks for suggestion! I have updated the code to align with padding values which were used on old design, please take a look. Each device resolution has its own specific values, which is why I had to define additional for mobile/tablet

acrobat/styles/styles.css
```diff
@@ -206,6 +205,7 @@ div.how-to {
     "image" !important;
   }
   .how-to-heading {
+    padding-top: 44px;
```
zagi25, line 208: This is overridden by https://github.com/adobecom/dc/pull/294/files#diff-b343378297f83896f1a341f07ea0505bbcfa331963369b08eeef401b34d49747R396 .

zagi25 (changes requested):
@saragajic11 This looks good but I just realized that this will affect all of our pages that uses `how-to` block, so you should target `how-to` only if it is inside `eventwrapper` : 

``` 
.eventwrapper  .how-to-heading { 
    ...
 }
```
etc.

zagi25 (approved):
If everyone is okay with extra padding on smaller devices, then this is fine by me. Great job @saragajic11 !

draganatrajkovic (approved):

PR #295, opened 2023-07-31
MWPW-131632

Blainegunn:
Resolves : [MWPW-13163](https://jira.corp.adobe.com/browse/MWPW-131632)

### DIFF
--- a/acrobat/blocks/dc-converter-widget/dc-converter-widget.js
+++ b/acrobat/blocks/dc-converter-widget/dc-converter-widget.js
@@ -121,6 +121,8 @@ const pageLang = localeMap[langFromPath] || 'en-us';
 export default function init(element) {
   element.closest('main > div').dataset.section = 'widget';
   const widget = element;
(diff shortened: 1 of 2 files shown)


acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -1,3 +1,5 @@
+import {createTag} from "../../scripts/miloUtils.js";
```
peller, line 1: this will make an extra network request at a critical time and slow down our performance

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -144,6 +151,13 @@ export default function init(element) {
   let REDIRECT_URL = '';
   let DC_GENERATE_CACHE_URL = '';
 
+  createTag.then((createTag) => {
```
peller, line 154: this is too late to prefetch. we need the prefetch on the initial HTML document.

PR #307, opened 2023-08-04
DCXY-18423 add child-src:blob on prod for adobe proxy F&S support on old Safari <15.4

peller:
to support frictionless fill & sign.  worker-src:blob is usually sufficient for them to create their wasm proxies, but for old Safari we determined that child-src is also needed


Blainegunn:
@peller we'll want this on dev/stage as well?

PR #308, opened 2023-08-06
replace hard-coded version with variable

peller:
dc-hosted/generate cache will always be the second half of the widget version tuple


Blainegunn (approved):
thank you

PR #315, opened 2023-08-10
MWPW-127299 use stage proxy dc urls

peller:
Test proxy urls on stage only. This will fetch DC widget assets from the same www.adobe.com host as the rest of the page and should eliminate delay.


peller:
@Blainegunn would like to see this on stage ASAP so we can test and then roll out to prod if successful

PR #325, opened 2023-08-16
MWPW-133659 l1 to l2 rotate pdf

draganatrajkovic:
Resolves: https://jira.corp.adobe.com/browse/MWPW-133659

**Llinks:**

Before: https://stage--dc--adobecom.hlx.page/acrobat/online/rotate-pdf
After: https://mwpw-133659-l1-to-l2-rotate-pdf--dc--adobecom.hlx.page/acrobat/online/rotate-pdf


acrobat/blocks/eventwrapper/eventwrapper.css
```diff
@@ -50,13 +50,27 @@ body:not([data-current-event]) [data-event-name=onload] {
...
   }
 }
+
+@media (max-width: 600px) {
+  .wapper-rotate-pages.fsw > div > div {
+    height: auto;
+  }
+}
+
+.wapper-rotate-pages .LifecycleOrganizeSettings__organizeSettings___2EChZ {
```
peller, line 69: I don't think it's safe to reference webpack-generated hash values from our build.   Are these style changes that we should be making in the widget?

Blainegunn: @draganatrajkovic if this is the best solution we can do something like this 
`div[id*="LifecycleOrganizeSettings__organizeSettings"]`
assuming it's a `div` :)

draganatrajkovic: @Blainegunn Thank you :)
@peller I fixed the code and small bugs, so please take a look.

acrobat/blocks/eventwrapper/eventwrapper.css
```diff
@@ -50,13 +50,27 @@ body:not([data-current-event]) [data-event-name=onload] {
...
   }
 }
+
+@media (max-width: 600px) {
+  .wapper-rotate-pages.fsw > div > div {
+    height: auto;
+  }
+}
+
+.wapper-rotate-pages [data-test-id="lifecycle-organize-fragment"] {
```
peller, line 69: not ideal to reference test ids since they're part of our test infrastructure. Can you do something similar with a class pattern match?

peller (approved):

acrobat/blocks/eventwrapper/eventwrapper.js
```diff
@@ -86,6 +92,38 @@ export default function init(element) {
       }
     }
 
+    if (verb === 'rotate-pages') {
```
peller, line 95: ok to do in a separate PR, but don't we need the same rules for the new fill & sign?  we'd also have to be careful to coordinate both rollouts

acrobat/blocks/eventwrapper/eventwrapper.css
draganatrajkovic: @peller Thank you for the comment, I fixed it. Also, I added some additional css to make it look good on smaller screens.

acrobat/blocks/eventwrapper/eventwrapper.js
draganatrajkovic: @peller when the widget for 'fill & sign' is ready, it will be easy to update the condition for 'rotate-pages' with `if (verb === 'rotate-pages' || verb === 'fillsign')` wherever applicable. That behaviour will be updated with this ticket https://jira.corp.adobe.com/browse/MWPW-133658.

Furthermore, in the last commit, I fixed the bug with styles in the case of a failing upload or when a corrupt file is uploaded.

draganatrajkovic:
@Blainegunn @peller 
Given the requests to remove the footer background-color and reposition the white part of the widget close to the top orange edge, I had to introduce a specific class for such pages. Thus, I added the class 'l2-state' to the <body> tag. This approach might not be conventional, which is why I wanted to verify with you if you have any concerns about it. I incorporated this change into the dc-converter-widget.js file to ensure the class is applied at the earliest possible moment.
 An alternative method using a query selector could cause delays, with changes taking effect after the page has already rendered. This could lead to late adjustments in elements' positions or styles, resulting in an unfavorable appearance. Particularly, the loading time of the small-footer might exacerbate this issue.
 
Once again, due to the need for some time to calculate the footer height and the requirement for the widget's height to be 100vh - (navHeight + footerHeight), certain content might experience displacement. In order to prevent a scenario like the one shown in the image at https://jira.corp.adobe.com/secure/attachment/10596793/10596793_image-2023-08-16-16-18-39-949.png , I expanded the orange background size. As a result, the footer will only slightly extend under the widget section (and perhaps in the production environment, where things are faster, this difference might not be as noticeable). However, this solution differs from the design, which anticipates that the orange background should occupy about half of the height as shown here:  https://xd.adobe.com/view/41201c56-4094-4482-8d51-caf5faa044c2-1473/screen/210ec4d2-bf4d-4d60-b991-fe66c317ff31/.

  The last request was to make this icon ticker https://jira.corp.adobe.com/secure/attachment/10596792/10596792_image-2023-08-16-16-17-54-257.png, and with this commit it is resolved.
  
 Please take a look, I need you approval so I can continue to work on responsive screens.

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -163,6 +163,10 @@ export default function init(element) {
 
   widget.querySelector('div').id = 'VERB';
   const VERB = widget.querySelector('div').textContent.trim().toLowerCase();
+  if (VERB === 'rotate-pages') {
+    const body = document.querySelector('body');
+    body.classList.add('l2-state');
```
peller, line 168: I like this feature-based approach better than citing the verb name directly in our CSS.  This should make it easier to migrate fill&sign styles as well.

peller (approved):

Blainegunn (approved):

zagi25 (approved):

draganatrajkovic:
@Blainegunn Last commit for responsive design is pushed

PR #335, opened 2023-08-21
Mwpw 127443 eliminate bowser

zagi25:
MILO team didn't want this functionality (getBrowserData) in their codebase, so I added it to ours and removed Bowser.js.

Resolves: [MWPW-127443](https://jira.corp.adobe.com/browse/MWPW-127443)

Before: https://stage--dc--adobecom.hlx.live/acrobat/online/pdf-to-jpg
After: https://mwpw-127443-eliminate-bowser--dc--adobecom.hlx.live/acrobat/online/pdf-to-jpg


acrobat/scripts/legacyBrowser.js
```diff
@@ -1,52 +1,22 @@
...
-
-  if (/Firefox/i.test(browserName)) return 'FF';
-
-  if (/Safari/i.test(browserName)) {
-    // if we cannot determine major version, we should not redirect to be on the safe side
-    // Safari: DC Widget > 12, Milo >= 14
-    return (!majorVersion || majorVersion >= 14) ? 'SAFARI' : 'SAFARI-LEGACY';
-  }
-  return null;
+  if (/Internet Explorer/i.test(browserName)) return true;
```
peller, line 8: for IE and old Edge, can we just use an expression like `/MSIE \d|Trident.*rv:/.test(navigator.userAgent)`?  There's no need to worry about future-proofing such checks.

acrobat/scripts/legacyBrowser.js
```diff
@@ -1,52 +1,22 @@
...
+  // EDGE: DC Widget >= 79, Milo >= 86
+  if (/Microsoft Edge/i.test(browserName) && majorVersion < 86) return true;
+  // Safari: DC Widget > 12, Milo >= 14
+  if (/Safari/i.test(browserName) && majorVersion < 14) return true;
```
peller, line 12: we don't check for old versions of other browsers with very small market share, do we need to do this for Safari?

acrobat/scripts/legacyBrowser.js
peller: Do we know how many users are hitting the EOL page?  Do we need to support this indefinitely?

acrobat/scripts/utils.js
```diff
@@ -35,3 +35,56 @@ export const [setLibs, getLibs] = (() => {
     }, () => libs,
   ];
 })();
+
+export const getBrowserData = () => {
```
peller, line 39: wondering if we need to break down browser by name or if we can rely on simpler logic (see other comments)

acrobat/blocks/eventwrapper/eventwrapper.js
```diff
@@ -59,12 +54,11 @@ export default function init(element) {
...
 
   const handleEvents = (e, jobData, converter, verb) => {
-    let parser = bowser.getParser(window.navigator.userAgent);
-    let browserName = parser.getBrowserName();
+    const { name: browserName, isMobile } = getBrowserData();
     let extID;
     if (e === PROCESS_START) converterAnalytics();
-    if (e === CONVERSION_COM && parser.parsedResult.platform.type === 'desktop'
-        || e === PREVIEW_DIS && parser.parsedResult.platform.type === 'desktop') {
+    if (e === CONVERSION_COM && !isMobile
```
peller, line 60: for the chrome extension, have we considered using [user agent client hints](https://web.dev/migrate-to-ua-ch/)?  It's not supported on all browsers, but it should be supported on the browsers we need (Chrome, new Edge)

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -1,3 +1,4 @@
+import redirectLegacyBrowsers from '../../scripts/legacyBrowser.js';
```
peller, line 1: at a minimum, consider loading this dynamically much later at a point which could not hurt our page load performance.  Alternatively, let's look for a simpler solution or perhaps even retiring the redirect if we can get PM to approve.

zagi25:
@peller, @Blainegunn maybe instead of parsing the userAgent we can do a feature check, this was proposed by the MILO team, as more reliable way to check for browser version, but I couldn't find the feature that corresponds to our needs manually at that time, so I thought that we need to do this this way. 

With the little help from the Browserstack just now ( I forgot that this exists, my bad) I saw in the console that code breaks on old browsers because MILO is using [replaceChildern](https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren) and this feature perfectly matches the Edge and Safari requirements (>= 86 && >= 14 -> https://caniuse.com/?search=replaceChildren).

So maybe we can ditch the whole parsing of userAgent, `try catch` this feature, and if it fails then we redirect? 

And we can use `userAgentData` for checking the extension.

peller:
@zagi25 why was this closed? Is there a new approach?

peller:
ok I had them mixed up. new PR is here for reference #337

PR #358, opened 2023-09-01
MWPW-127443

Blainegunn:
Re-Resolves : [MWPW-127443](https://jira.corp.adobe.com/browse/MWPW-127443)
Sadly `userAgentData` doesn't work on some of the legacy browser we target
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData

### DIFF
--- a/acrobat/blocks/dc-converter-widget/dc-converter-widget.js
+++ b/acrobat/blocks/dc-converter-widget/dc-converter-widget.js
@@ -184,6 +184,15 @@ export default async function init(element) {
     REDIRECT_URL_DIV.remove();
   }
 
+  // Feature checking for old browsers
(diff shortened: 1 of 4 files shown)


peller:
> Re-Resolves : [MWPW-127443](https://jira.corp.adobe.com/browse/MWPW-127443) Sadly `userAgentData` doesn't work on some of the legacy browser we target https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData

Right. My understanding was that we were going to use `userAgentData` only for the purpose of detecting new Chrome and new Edge for the extension popups.

For EOL detection, I thought we were going to use feature detection to look for something that wasn't supported in those old browsers (replaceChildren?) so that userAgent parsing would no longer be necessary?

PR #361, opened 2023-09-01
fix(csp): enable RUM reporting

trieloff:
We've been seeing CSP violations due to Franklin RUM requests being blocked. This commit allow-lists rum.hlx.page which is our RUM collection endpoint

### DIFF
--- a/acrobat/scripts/contentSecurityPolicy/prod.js
+++ b/acrobat/scripts/contentSecurityPolicy/prod.js
@@ -45,6 +45,7 @@ const connectSrc = [
   '*.typekit.net/',
   'trial-eum-clienttons-s.akamaihd.net/',
   '*.akstat.io/',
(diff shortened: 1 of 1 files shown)


davidnuescheler:
maybe related.... but i saw that the `.page` and the `.live` domains were also part of the CSP? i think that's wrong and potentially dangerous as there shouldn't be any traffic sent from acom to either `.page` or `.live`?

Blainegunn:
@davidnuescheler we don't send any traffic to `.page` or `.live` they were in the CSP when we got some resources from them in the early days. I'll remove them now as they serve no purpose. 
@trieloff thanks for opening this.

Blainegunn:
For testing `?rum=on`
https://www.adobe.com/acrobat/online/sign-pdf.html?rum=on

zagi25 (approved):

PR #379, opened 2023-09-18
MWPW-136128: Added star-outline icon

zagi25:
Resolves: [MWPW-136128](https://jira.corp.adobe.com/browse/MWPW-136128)

Before: https://stage--dc--adobecom.hlx.live/acrobat/online/rotate-pdf
After: https://mwpw-136128-star-rating-rotate--dc--adobecom.hlx.live/acrobat/online/rotate-pdf


zagi25:
@Blainegunn, @saragajic11 , @puruadobe  I think we should create a separate ticket for fixing stylelint errors because there is a lot of them , and also we should probably change some of the lint rules .

Blainegunn (approved):
but could you fix the linting issues please? i know they aren't yours but just as a courtesy 🙏

saragajic11 (approved):

PR #397, opened 2023-10-03
MWPW-137308: Added origin-trial meta tag so scheduler.yield() can be used

zagi25:
Resolves: [MWPW-137308](https://jira.corp.adobe.com/browse/MWPW-137308)

Before: https://stage--dc--adobecom.hlx.live/acrobat/online/pdf-to-ppt
After: https://mwpw-137308-scheduler-yield--dc--adobecom.hlx.live/acrobat/online/pdf-to-ppt

### DIFF
--- a/acrobat/scripts/scripts.js
+++ b/acrobat/scripts/scripts.js
@@ -46,6 +46,14 @@ const getLocale = (locales, pathname = window.location.pathname) => {
   return locale;
 };
 
+function getTrialToken() {
+  const LOCAL_TOKEN = 'AloskvzI3ilSEcG1AXxQ9rv0Ro12O3ISCc3cZI/5yRr2mi2Eap5t0N5L3OT8KpWRBBOnfDvntk8n+dCG4eHGcQsAAABReyJvcmlnaW4iOiJodHRwOi8vbG9jYWxob3N0OjMwMDAiLCJmZWF0dXJlIjoiU2NoZWR1bGVyWWllbGQiLCJleHBpcnkiOjE3MDk2ODMxOTl9';
+  const PROD_TOKEN = 'ApPnSNHCIWK27DqNdhiDHtOnC8mmBgtVJX5CLfG0qKTYvEG3MRpIdFTlz35GPStZLs926t+yC9M4Y6Ent+YKbgkAAABkeyJvcmlnaW4iOiJodHRwczovL2Fkb2JlLmNvbTo0NDMiLCJmZWF0dXJlIjoiU2NoZWR1bGVyWWllbGQiLCJleHBpcnkiOjE3MDk2ODMxOTksImlzU3ViZG9tYWluIjp0cnVlfQ==';
+  const { hostname } = window.location;
+  if (hostname.includes('localhost')) return LOCAL_TOKEN;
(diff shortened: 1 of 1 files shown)


acrobat/scripts/scripts.js
```diff
@@ -101,6 +109,12 @@ const getBrowserData = (userAgent) => {
 // Get browser data
 window.browser = getBrowserData(window.navigator.userAgent);
 
+// Add origin-trial meta tag
+const tokenElement = document.createElement('meta');
+tokenElement.httpEquiv = 'origin-trial';
+tokenElement.content = getTrialToken();
+document.head.appendChild(tokenElement);
```
peller, line 116:
```suggestion
const token = getTrialToken();
if (token) {
  const tokenElement = document.createElement('meta');
  tokenElement.httpEquiv = 'origin-trial';
  tokenElement.content = token;
  document.head.appendChild(tokenElement);
}
```

acrobat/scripts/scripts.js
```diff
@@ -46,6 +46,14 @@ const getLocale = (locales, pathname = window.location.pathname) => {
   return locale;
 };
 
+function getTrialToken() {
+  const LOCAL_TOKEN = 'AloskvzI3ilSEcG1AXxQ9rv0Ro12O3ISCc3cZI/5yRr2mi2Eap5t0N5L3OT8KpWRBBOnfDvntk8n+dCG4eHGcQsAAABReyJvcmlnaW4iOiJodHRwOi8vbG9jYWxob3N0OjMwMDAiLCJmZWF0dXJlIjoiU2NoZWR1bGVyWWllbGQiLCJleHBpcnkiOjE3MDk2ODMxOTl9';
+  const PROD_TOKEN = 'ApPnSNHCIWK27DqNdhiDHtOnC8mmBgtVJX5CLfG0qKTYvEG3MRpIdFTlz35GPStZLs926t+yC9M4Y6Ent+YKbgkAAABkeyJvcmlnaW4iOiJodHRwczovL2Fkb2JlLmNvbTo0NDMiLCJmZWF0dXJlIjoiU2NoZWR1bGVyWWllbGQiLCJleHBpcnkiOjE3MDk2ODMxOTksImlzU3ViZG9tYWluIjp0cnVlfQ==';
+  const { hostname } = window.location;
+  if (hostname.includes('localhost')) return LOCAL_TOKEN;
+  if (hostname === 'www.stage.adobe.com' || hostname === 'www.adobe.com') return PROD_TOKEN;
```
peller, line 54: if endsWith(adobe.com) return PROD_TOKEN else return undefined

peller: or simplify and maybe just unconditionally post the meta tag.  The meta tag should be a no-op when running on localhost, so scheduler.yield just won't be available, as would be the case for non-Chrome browsers.  The widget code should tolerate this.

Blainegunn:
@zagi25 now that you have access to the headers sheet you want to move this there?

Blainegunn (changes requested):
@zagi25 now that you have access to the headers sheet you want to move this there?

zagi25:
As @Blainegunn said we can move the token to the headers doc, so it will be delivered as response header
Screenshot 2023-10-06 at 12 46 36

There is no need for any code update.
cc: @peller

PR #413, opened 2023-10-12
MWPW-137590: add loc html

nishantka:
- Adding html files for localization in dc

Currently, this change will be on `loc` branch
Resolves: [MWPW-137590](https://jira.corp.adobe.com/browse/MWPW-137590)


Blainegunn (approved):
@nishantka thanks for hooking us up!

sukamat (approved):

PR #417, opened 2023-10-13
MWPW-137930 DCXY-18989 if snippet is loaded inline, use it

peller:
The edge worker will load the dc-generate-cache snippet and inline it in the HTML page in a section tag id="edge-snippet".  If it's already there, use the DOM rather than fetch it over http.


acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -317,7 +314,14 @@ export default async function init(element) {
   const verbIncludeList = ['compress-pdf', 'fillsign', 'sendforsignature', 'add-comment',
     'delete-pages', 'reorder-pages', 'split-pdf', 'insert-pdf', 'extract-pages', 'crop-pages', 'number-pages'];
 
-  if (verbIncludeList.includes(VERB) || preRenderDropZone) {
+  const INLINE_SNIPPET = widget.querySelector(':scope > section#edge-snippet');
+  if (INLINE_SNIPPET) {
+    widgetContainer.dataset.rendered = 'true';
```
JFernandezAdobe, line 319: Hi @peller, we probably need a unit test for lines 319-323?

Blainegunn (approved):
looks fine to me

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
peller: done

PR #418, opened 2023-10-16
MWPW-137924: Updated csp rules

zagi25:
Resolves: [MWPW-137924](https://jira.corp.adobe.com/browse/MWPW-137924)

Before: https://stage--dc--adobecom.hlx.live/acrobat/online/jpg-to-pdf
After:  https://mwpw-137924-widget-fails-to-load--dc--adobecom.hlx.live/acrobat/online/jpg-to-pdf


peller (approved):

peller (changes requested):
this is the wrong url and it's a 404.  need to find a fix, might be an issue with the widget

PR #448, opened 2023-11-03
MWPW-138677 Reorg E2E test folder

TsayAdobe:
- Remove `frictionless` folder
- Update the readme and workflow
- Use `stage` branch for E2E tests in GitHub Actions
- Port in @JFernandezAdobe's E2E fix for [the Milo PR](https://github.com/adobecom/milo/pull/1474)


puruadobe (approved):

.github/workflows/test-milo.yml
```diff
@@ -2,6 +2,12 @@ name: Verify Milo Changes
 
 on:
   workflow_dispatch:
+    inputs:
+      testCodeBranch:
+        description: 'Test Code Branch'
+        required: true
```
JFernandezAdobe, line 8: Does this input need to be supplied by https://github.com/adobecom/milo/blob/main/.github/workflows/dispatch.yml?

JFernandezAdobe (approved):

.github/workflows/test-milo.yml
TsayAdobe: Examples from GitHub have `required: true` with a default value. I think it will use the default value if it is not supplied.

PR #462, opened 2023-11-16
MWPW-138331: Fixed canonical url bug

zagi25:
https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L301
I overlooked this line, pathname for sign.ing is `/` and canonical url pathname is `/acrobat/online/sign-pdf`, so adding `useDotHtml: true` doesn't work.

Resolves: [MWPW-138331](https://jira.corp.adobe.com/browse/MWPW-138331)

Before: https://stage--dc--adobecom.hlx.live/acrobat/online/sign-pdf
After: https://mwpw-138331-canonical-url-ing-pages--dc--adobecom.hlx.live/acrobat/online/sign-pdf

### DIFF
--- a/acrobat/scripts/scripts.js
+++ b/acrobat/scripts/scripts.js
@@ -111,6 +111,12 @@ if (hostname === 'www.stage.adobe.com') {
   document.head.appendChild(tokenElement);
 }
 
+// Adding .html to canonical url for .ing pages
+if (hostname.endsWith('.ing')) {
+  const canonEl = document.head.querySelector('link[rel="canonical"]');
+  canonEl?.setAttribute('href', `${canonEl.href}.html`);
+}
+
 function loadStyles(paths) {
   paths.forEach((path) => {
     const link = document.createElement('link');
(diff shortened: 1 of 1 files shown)


seanchoi-dev (approved):

acrobat/scripts/scripts.js
```diff
@@ -111,6 +111,12 @@ if (hostname === 'www.stage.adobe.com') {
   document.head.appendChild(tokenElement);
 }
 
+// Adding .html to canonical url for .ing pages
+if (hostname.endsWith('.ing')) {
+  const canonEl = document.head.querySelector('link[rel="canonical"]');
+  canonEl?.setAttribute('href', `${canonEl.href}.html`);
```
seanchoi-dev, line 117: What if there is `.html` already?
I checked https://edit.ing/acrobat/online/sign-pdf and it seems there is already `.html`
image

seanchoi-dev (commented):
What if there is `.html` already?

acrobat/scripts/scripts.js
zagi25: Yeah, you're right, just in case I need to check if it ends with `.html`. On that page there is `.html` because my previous solution works on that link.

TsayAdobe (approved):

PR #489, opened 2023-11-21
DCXY-20504: Don't pre-render dropzone when l2 limit is hit and upsell is shown

kbenelli-adobe:
…is shown.

PR #490, opened 2023-11-22
DCXY-19397: Don't pre-render dropzone when l2 limit is hit and upsell is shown

peller:
DC Web will keep a set of cookies for when a limit has been reached. We can check this cookie to predict whether an upsell will be shown and only pre-render the dropzone if that cookie is not set.

merged from [kbenelli-adobe:DCXY-19397-New](https://github.com/kbenelli-adobe/dc/tree/DCXY-19397-New)

@kbenelli-adobe


peller:
closing in favor of #489

PR #492, opened 2023-11-30
Updated libs for stage hostname

saragajic11:
Resolves: [[MWPW-139835](https://jira.corp.adobe.com/browse/MWPW-139835)]


draganatrajkovic (approved):

peller:
ok, I just updated the edgeworker so that when it inlines libs/styles/styles.css it grabs it from stage.adobe.com instead of adobe.com

TsayAdobe (changes requested):
Do we use `setLibs()` in `utils.js`?

We should unit test `setLibs()` in `scripts.js`? There are too many `if`conditions. We should cover all condition branches.

PR #508, opened 2023-12-15
Fix DC converter widget block unit tests

TsayAdobe:
Add more tests and do a cleanup


test/blocks/dc-converter-widget/mocks/body_gen_cache.html
```diff
@@ -0,0 +1,16 @@
+<main>
+    <div class="dc-converter-widget">
+        <div>
+            <div>pdf-to-ppt</div>
+        </div>
+        <div>
+            <div>https://www.adobe.com/go/testredirect</div>
+        </div>    
+        <div>
+            <div>https://www.adobe.com/generate_cache_url</div>
```
peller, line 10: are we even using this?

TsayAdobe: @peller We can revisit the design. If the feature is not used and will never be used, we should remove the related code.

Blainegunn (approved):
ty @TsayAdobe

JFernandezAdobe (approved):

PR #514, opened 2024-01-10
MWPW 138749 Visual tests for merch cards

TsayAdobe:
- Add the visual tests into integration tests
- Add a test job into the workflow triggered when a PR is merged to Milo main. 
- Test Chromium screenshots only in the workflow. There are high variations in rendering for Firefox and WebKit's. Looking for solutions.
- The baseline screenshots are from macOS 12.7.2, which is the latest macOS in GitHub Actions. The workflow runs on macOS only. 
- Accordion tests and screenshots are updated.


JFernandezAdobe (approved):

JFernandezAdobe:
Looks great @TsayAdobe!

PR #528, opened 2024-01-29
CSP Update

dc-csp-update:


peller (changes requested):
Needs review.  Which of these hosts are referenced by our code, and should they be referenced by our code?  kaspersky-labs very likely is client-side software the user has installed, it should not be allowed from our page.

acrobat/scripts/contentSecurityPolicy/dev.js
```diff
@@ -58,6 +58,26 @@ const connectSrc = [
...
+  'accounts.google.com',
+  'api.echosign.com',
+  'c.clicktale.net',
+  'ff.kes.v2.scr.kaspersky-labs.com',
+  'ff.kis.v2.scr.kaspersky-labs.com',
+  'files.acrobat.com',
+  'gc.kes.v2.scr.kaspersky-labs.com',
+  'gc.kis.v2.scr.kaspersky-labs.com',
+  'me.kes.v2.scr.kaspersky-labs.com',
+  'me.kis.v2.scr.kaspersky-labs.com',
```
peller, line 80: not sure about all of the changes, but at least the kaspersky-labs ones are a hard no.  if we're just going by logged CSP failures, some of them (many of them) may be blocked for good reason.

peller (changes requested):

PR #531, opened 2024-02-01
eventwrapper fix on bigger screen resolutions for FnS and rotate-pdf …

saragajic11:
Resolves: [MWPW-141570](https://jira.corp.adobe.com/browse/MWPW-141570)


Blainegunn:
merging to allow testing

PR #554, opened 2024-02-16
MWPW-142806

Blainegunn:
Resolves : [MWPW-142806](https://jira.corp.adobe.com/browse/MWPW-142806)

### DIFF
--- a/acrobat/scripts/scripts.js
+++ b/acrobat/scripts/scripts.js
@@ -275,6 +275,7 @@ const CONFIG = {
     version: '1.0',
     onDemand: false,
   },
+  imsScope: 'AdobeID,openid,gnav,firefly_api,adobeid.scope',
 };
 
 // Setting alternative Jarcis client ID for these paths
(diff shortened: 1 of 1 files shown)


acrobat/scripts/scripts.js
```diff
@@ -275,14 +275,15 @@ const CONFIG = {
     version: '1.0',
     onDemand: false,
   },
+  imsScope: 'AdobeID,openid,gnav,firefly_api,adobeid.scope',
```
narcis-radu, line 278: Hey Blaine! Milo-college projects will inherit the default scope from Milo as mentioned in the [wiki](https://wiki.corp.adobe.com/display/adobedotcom/Universal+Navigation+integration+guide#UniversalNavigationintegrationguide-IMSlibadditionalscopeconfiguration), my plan is to add the `pps.read` and `firefly_api` scopes as soon as I'm done with validating that all IMS Client IDs have the scopes approved.
If you want to override the scopes, that's fine, but you will need to add `pps.read` and `firefly_api`. However, I think it's better if you get them from Milo, this way you're always in sync.

PR #557, opened 2024-02-21
Fix Libs config to match MILO 

Blainegunn:
Fix Libs config to match MILO


TsayAdobe:
Please review the test cases in `scripts.jest.js` and `utils.jest.js`.

test/scripts/scripts.jest.js
```diff
@@ -118,7 +118,7 @@ describe('Test scripts', () => {
       await require('../../acrobat/scripts/scripts.js');
       await delay(100);
       const config = await getConfig();
-      expect(config.miloLibs).toEqual('http://localhost:6456/libs');
+      expect(config.miloLibs).toEqual('/libs');
```
TsayAdobe, line 121: When `milolibs=local`, it should be  `http://localhost:6456/libs`

test/scripts/utils.jest.js
```diff
@@ -34,7 +34,7 @@ describe('Test utils.js', () => {
       'https://www.stage.adobe.com?milolibs=local'
     );    
     const libs = setLibs('/libs');
-    expect(libs).toBe('http://localhost:6456/libs');
+    expect(libs).toBe('/libs');
```
TsayAdobe, line 37: When `milolibs=local`, it should be `http://localhost:6456/libs`

TsayAdobe (changes requested):
We need to enable devs using a local milo.

joaquinrivero (approved):

PR #558, opened 2024-02-27
Update EdgeWorker code

JFernandezAdobe:
Updated with version 0.83 from stg

Related ticket: [MWPW-141047](https://jira.corp.adobe.com/browse/MWPW-141047)

https://mwpw-141047-update-ew--dc--adobecom.hlx.live/acrobat/online/jpg-to-pdf?martech=off


peller (approved):

TsayAdobe (approved):
If prod and stage have the same code, we don't need this line in `prod2stg.js`
```
mainjs = mainjs.replace(/isProd = true/g, 'isProd = false');
```
In fact, we can deploy the same folder to Akamai. Anyway, we can refine it later.

PR #604, opened 2024-03-15
Fix E2E tests

TsayAdobe:
- Sign PDF is changed to L1. Remove Sign PDF test cases and add it to L1.
- Pricing checking step may take longer than 1 min
- Test cases MWPW-137251 and MWPW-137439 are set inactive in Jira.


JFernandezAdobe (approved):

Blainegunn (approved):
🙏

PR #612, opened 2024-03-25
Mwpw 145314

Blainegunn:


acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -111,6 +111,7 @@ const verbRedirMap = {
   'compress-pdf': 'compress',
   'png-to-pdf': 'jpgtopdf',
   'number-pages': 'number',
+  'ocr-pdf': 'createpdf',
```
peller, line 114: we'll need to make this redirect to [ocr-pdf](https://acrobat.adobe.com/link/acrobat/ocr-pdf) via a new go url, correct?

Blainegunn: i asked you guys for GoUrl and I believe you said there wasn't one so it'll go to the default until we are given one

peller: I found April's wiki and added an entry here: https://wiki.corp.adobe.com/display/DEX/DC+Web+App2+rollout 
Now we just need someone to create the go urls.

PR #616, opened 2024-03-26
Temporarily remove gnav smoke test

TsayAdobe:
Currently the new universal nav on some pages has "Free Trial" in a CTA and "Buy now" a menu item. Some pages the other way around. Temporarily remove the test until it is fixed.


TsayAdobe:
Not needed. The [PR](https://github.com/adobecom/dc/pull/620) should fix it.

PR #625, opened 2024-04-03
MWPW-145718 Enable running WTR on different Milo branches

TsayAdobe:
* Use [Rollup replace plugin](https://github.com/rollup/plugins) to dynamically update the client scripts, which include production code and Web Test Runner test cases.
* The branch name is passed in through the environment variable `MILOLIBS`. e.g. use the command lines
   ```MILOLIBS=stage npm test``` 
   ```MILOLIBS=main–milo–tsayadobe npm test```
   ```MILOLIBS=local npm test```
If no environment variable is set, tests will use the original scripts, which use the main branch of Milo
* This PR is for Web Test Runner. It doesn't work for Jest tests.
* Caveats: 
  * There could be a false pass if the env variable is not set correctly. e.g. a typo in the environment variable name then tests running on the Milo main branch.
  *  Strings in `expect` could be replaced if it matches the text to be replaced. If there is an unexpected result, do debugging in a browser and review scripts in the sources.  

Test URLs:
N/A


Blainegunn (approved):

Blainegunn:
@TsayAdobe a few conflicts

joaquinrivero (approved):

PR #635, opened 2024-04-15
MWPW-146462

Blainegunn:
Resolves: [MWPW-146462](https://jira.corp.adobe.com/browse/MWPW-146462)


Blainegunn:
basically trivial so i'm merging so QA can make sure there's no regression

PR #640, opened 2024-04-22
MWPW-142133 Prevent reentry of PDF widget init

TsayAdobe:
- There is always one and only one PDF widget in a PDF frictionless page
- Prevent
  - two dc-converter-widget blocks in a page
  -  init() is called more than once
  - Other mistakes in an EdgeWorker   

Test URLs:

Before: https://main--dc--adobecom.hlx.live/acrobat/online/word-to-pdf?martech=off
After: https://mwpw-142133--dc--adobecom.hlx.live/acrobat/online/word-to-pdf?martech=off


Blainegunn (approved):

joaquinrivero (approved):

acrobat/blocks/dc-converter-widget/dc-converter-widget.js
```diff
@@ -135,6 +135,7 @@ const langFromPath = url.pathname.split('/')[1];
 const pageLang = localeMap[langFromPath] || 'en-us';
 
 export default async function init(element) {
+  if (document.querySelector('div[data-section="widget"]')) return;
```
peller, line 138: duplication was happening at the edge worker. Not sure this will help, but probably can't hurt either?

TsayAdobe: Do you mean what we saw was two `<section id="edge-snippet">`s? Anyway, the change is to safeguard `init()` so it won't be called twice or more.

peller: Yes, I believe so.

PR #644, opened 2024-04-26
MWPW-147046

Blainegunn:
Resolves : [MWPW-147046](https://jira.corp.adobe.com/browse/MWPW-147046)
@peller @kbenelli-adobe do we need to a [value](https://github.com/adobecom/dc/blob/stage/acrobat/blocks/dc-converter-widget/dc-converter-widget.js#L117) for `exhLimitCookieMap` for `chat-pdf`.


acrobat/styles/styles.css
```diff
@@ -23,6 +23,20 @@ div[data-section="widget"] {
   min-height: 570px;
 }
 
+#dc-converter-widget[verb=chat-pdf] {
+  background-color: #fff;
+  background-image: url(../img/backgrounds/chat-pdf.png);
```
peller, line 28: I suggested to the designers that we use a gradient instead. not sure whether loading a background image would affect our performance, but we can tweak this later

peller (approved):

peller:
For the chat-pdf verb, we aren't predicting limit exhaustion in the cookie.  The only limits are the abuse limits. So I think what we want is just to not add anything to the map, let it assume the limit is not exhausted and paint the dropzone.  Sound good @kbenelli-adobe?

saragajic11 (approved):

joaquinrivero (approved):

acrobat/scripts/scripts.js
```diff
@@ -321,6 +321,7 @@ const { ietf } = getLocale(locales);
     const blockName = widgetBlock.classList.value;
     widgetBlock.removeAttribute('class');
     widgetBlock.id = 'dc-converter-widget';
+    widgetBlock.setAttribute('verb', verb);
```
peller, line 324: should we have used data-verb maybe?

acrobat/styles/styles.css
```diff
@@ -23,6 +23,20 @@ div[data-section="widget"] {
...
+  background-color: #fff;
+  background-image: url('../img/backgrounds/chat-pdf.png');
+  background-position: top left;
+  min-height: 623px;
+  background-repeat: no-repeat;
+  background-size: contain; 
+}
+
+/* DC Widget Override */
+div [class *= PdfnowLifecycle__dropzone2___] {
```
peller, line 36: hang on. don't we want this scoped for verb=chat-pdf only?  Also, can we use a more efficient selector like `^=` ?

PR #658, opened 2024-05-14
MWPW-144821: Load media from content root

joaquinrivero:
Resolves: [MWPW-144821](https://jira.corp.adobe.com/browse/MWPW-144821)

Updates elements with paths starting with ./media_ to use the base path defined in CONFIG.contentRoot

Test URLs
Before: https://main--dc--adobecom.hlx.page/documentcloud
After: https://mwpw-144821--dc--adobecom.hlx.page/documentcloud


saragajic11 (approved):

acrobat/scripts/scripts.js
```diff
@@ -290,6 +290,23 @@ const CONFIG = {
   imsScope: 'AdobeID,openid,gnav,pps.read,firefly_api,additional_info.roles,read_organizations',
 };
 
+function replaceDotMedia(area = document) {
+  // eslint-disable-next-line compat/compat
+  const currUrl = new URL(window.location);
+  const pathSeg = currUrl.pathname.split('/').length;
+  if (pathSeg >= 3) return;
```
TsayAdobe, line 297: Why do we need this? I don't see it in CC's.

acrobat/scripts/scripts.js
```diff
@@ -290,6 +290,23 @@ const CONFIG = {
...
+  const resetAttributeBase = (tag, attr) => {
+    area.querySelectorAll(`${tag}[${attr}^="./media_"]`).forEach((el) => {
+      // eslint-disable-next-line compat/compat
+      el[attr] = `${new URL(`${CONFIG.contentRoot}${el.getAttribute(attr).substring(1)}`, window.location).href}`;
```
TsayAdobe, line 301: The most important line doesn't seem to be in the code coverage.

acrobat/scripts/scripts.js
joaquinrivero: They do have it on their code to run it just when is on root, here is the code:

https://github.com/adobecom/cc/blob/stage/creativecloud/scripts/utils.js#L61-L72

TsayAdobe (approved):

PR #700, opened 2024-07-08
mwpw-153662 removed special characters from geo ip phone num

saragajic11:
Resolves: [MWPW-153662](https://jira.corp.adobe.com/browse/MWPW-153662)
Before: https://main--dc--adobecom.hlx.page/hk_en/acrobat/business/integrations?akamaiLocale=HK_EN
After: https://mwpw-153662-remove-special-chars--dc--adobecom.hlx.page/hk_en/acrobat/business/integrations?akamaiLocale=HK_EN


TsayAdobe (commented):
This should be a content issue. If authors fix the content, we don't need to run the find-replace code on all visitors.

saragajic11:
> This should be a content issue. If authors fix the content, we don't need to run the find-replace code on all visitors.

It can be an authoring fix, but I thought it might not be a bad idea to prevent issues like this from happening. It's a minor fix anyways

TsayAdobe (approved):

joaquinrivero (approved):

TsayAdobe (approved):

PR #719, opened 2024-07-22
MWPW-147913: Prevent Users From Uploading When They're Signed In

joaquinrivero:
Resolves: [MWPW-147913](https://jira.corp.adobe.com/browse/MWPW-147913)

Test url:

https://main--dc--adobecom.hlx.page/acrobat/online/compress-pdf?martech=off
https://mwpw-147913--dc--adobecom.hlx.page/acrobat/online/compress-pdf?martech=off


Blainegunn (changes requested):
looks like we're dispatching `IMS:Ready` twice in this file. 
Perhaps we should consolidate.
Maybe was can do `dropZoneContent.style.pointerEvents = 'none';` with `.css` and then remove it after IMS loads?

joaquinrivero (commented):
@Blainegunn I will find another way to get it properly

saragajic11 (approved):

peller:
❗  this covers a corner case but could degrade performance significantly for all users and may miss new cases for error handling

acrobat/scripts/scripts.js
```diff
@@ -338,6 +338,10 @@ if (IMS_GUEST) {
   };
 }
 
+// Load IMS
+const { default: dcLoadIMS } = await import('./dcIMS.js');
```
peller, line 342: also FYI this import adds delay to the startup, so we generally try to inline code here or make the inlining happen in the edgeworker as needed

PR #725, opened 2024-07-28
Update config.yaml

joaquinrivero:
Ignoring Kodiak warns, this is for internal use only


joaquinrivero:
Wrong fork, this should stay on dc-sandbox

PR #741, opened 2024-08-03
Call IMS Sooner

Blainegunn:
To help the page performance of the IMS Guest Token UX


acrobat/scripts/scripts.js
```diff
@@ -455,6 +455,12 @@ const { ietf } = getLocale(locales);
   const { loadArea, setConfig, loadLana, getMetadata, loadIms } = await import(`${miloLibs}/utils/utils.js`);
   addLocale(ietf);
 
+  // IMS Ready
+  loadIms().then(() => {
```
peller, line 459: do we need to call `setConfig` before `loadIMS`?  does loadIMS rely on the client id, etc?

PR #747, opened 2024-08-06
Turn on RUM collection from widget

peller:
track performance for ims_guest rollout to compare widget loading time and time to interactivity


TsayAdobe:
Should we add a condition for the stage, or ensure it doesn't reach production?

TsayAdobe (approved):

peller:
> Should we add a condition for the stage, or ensure it doesn't reach production?

It's fine for this to go to production as well.

PR #770, opened 2024-08-20
DCXY-25858 don't use ims token or pre-render for upsell redirections from extension

peller:


peller:
awaiting confirmation from the extensions team

peller:
Confirmed. @Blainegunn could you please take this to stage today?  The extensions team in Noida will test and if all goes well we can promote to prod before the restricted period begins.  Thanks!

Blainegunn (approved):

PR #784, opened 2024-09-04
implement integration with project unity

saragajic11:
Resolves: [MWPW-155563](https://jira.corp.adobe.com/browse/MWPW-155563)
Before: https://main--dc--adobecom.hlx.page/acrobat/online/test/unity-test?unitylibs=stage&martech=off
After: https://mwpw-155563-integrate-with-project-unity--dc--adobecom.hlx.page/acrobat/online/test/unity-test?unitylibs=stage&martech=off

### DIFF
--- a/acrobat/blocks/unity/unity.js
+++ b/acrobat/blocks/unity/unity.js
@@ -19,5 +19,5 @@ export default async function init(el) {
   });
   await stylePromise;
   const { default: wfinit } = await import(`${unitylibs}/core/workflow/workflow.js`);
-  await wfinit(el, 'cc', unitylibs);
+  await wfinit(el, 'acrobat', unitylibs);
 }
--- a/acrobat/blocks/unity/unity.js
+++ b/acrobat/blocks/unity/unity.js
@@ -1,0 +1,23 @@
+import { loadStyle } from '../../scripts/utils.js';
+
+function getUnityLibs(prodLibs = '/unitylibs') {
+  const { hostname } = window.location;
+  if (!hostname.includes('hlx.page')
+    && !hostname.includes('hlx.live')
+    && !hostname.includes('localhost')) {
+    return prodLibs;
+  }
+  const branch = new URLSearchParams(window.location.search).get('unitylibs') || 'main';
+  if (branch.indexOf('--') > -1) return `https://${branch}.hlx.live/unitylibs`;
+  return `https://${branch}--unity--adobecom.hlx.live/unitylibs`;
+}
+
+export default async function init(el) {
+  const unitylibs = getUnityLibs();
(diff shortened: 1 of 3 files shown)


acrobat/scripts/utils.js
```diff
@@ -35,3 +35,7 @@ export const [setLibs, getLibs] = (() => {
     }, () => libs,
   ];
 })();
+
+const miloLibs = setLibs('/libs');
+const { loadStyle } = await import(`${miloLibs}/utils/utils.js`);
```
peller, line 40: won't this slow down the initial load?

acrobat/blocks/unity/unity.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export default async function init(el) {
+  const unitylibs = getUnityLibs();
+  const stylePromise = new Promise((resolve) => {
+    loadStyle(`${unitylibs}/core/styles/styles.css`, resolve);
+  });
+  await stylePromise;
+  const { default: wfinit } = await import(`${unitylibs}/core/workflow/workflow.js`);
+  await wfinit(el, 'acrobat', unitylibs);
```
Ruchika4, line 22: @saragajic11 please pass 'v2' as the last parameter of this function as acrobat will be using the new refactored code for loading unity.
We added the version parameter as unity block for photoshop is live using the old version for now.

cc @aishwaryamathuria @sanjayms01

Ruchika4 (changes requested):

acrobat/scripts/utils.js
aishwaryamathuria: @peller The DC scripts.js is not loading this utils.js but is already loading the milolibs utils.js so this second call would come from disk cache (if the disable cache is not explicitly disabled in deveoper tools)

peller: is this coupling necessary?

acrobat/blocks/unity/unity.js
saragajic11: @Ruchika4 done: https://github.com/adobecom/dc/pull/784/commits/4407a6c42469554c7d1429ea4a7580ff84ac6de2 
thanks!

Blainegunn (changes requested):
@saragajic11 one more change
please update `wfinit` to include a language and region code.
**Example**
language code `en`
region code `us`
```
await wfinit(el, 'acrobat', unitylibs, 'v2', languageRegion, languageCode);
```
the codes should match the values from the `localeMap` in `acrobat/blocks/dc-converter-widget/dc-converter-widget.js`

Blainegunn (approved):

Blainegunn (approved):

PR #785, opened 2024-09-05
MWPW-157573 Optimize EdgeWorker for Mobile Widget

TsayAdobe:
* Enable different caches for desktop, tablet, and mobile devices.
* If the metadata `mobile-widget` is true and the client device type is mobile, the DC converter widget is not required. Eliminate unnecessary preloads and embedded scripts.
* The change has been deployed to the stage environment.
 
Resolve: [MWPW-157573](https://jira.corp.adobe.com/browse/MWPW-157573)

Test URLs:
https://www.stage.adobe.com/acrobat/online/sign-pdf.html
https://www.stage.adobe.com/acrobat/online/pdf-editor.html


saragajic11 (approved):

edgeworkers/Acrobat_DC_web_prod/main.js
```diff
@@ -86,32 +88,33 @@ export async function responseProvider(request) {
...
-    const snippetBody = snippet.substring(snippet.indexOf('<body>')+6, snippet.indexOf('</body>'));
-
-    rewriter.onElement('head', el => {
-      el.append(snippetHead);
-    });
-    rewriter.onElement('div.dc-converter-widget', el => {
-      el.append(`<section id="edge-snippet">${snippetBody}</section>`);
-    });
-
+    if (!(mobileWidget && request.device.isMobile)) {
```
peller, line 97: might be easier to read w/o the extra parens.  Do we mean that both of these conditions must be false or just one?

edgeworkers/Acrobat_DC_web_prod/main.js
```diff
@@ -122,13 +125,20 @@ export async function responseProvider(request) {
...
-      .replace('import(\'../../scripts/frictionless.js\')', 'import(\'/acrobat/scripts/frictionless.js\')')
-    + scripts
-      .replace('const { default: dcConverter } = await import(`../blocks/${blockName}/${blockName}.js`);', '')
-      .replace('await import(\'./contentSecurityPolicy/csp.js\')', '{default:()=>{}}')
-      .replace('await import(\'./dcLana.js\')', 'await import(\'/acrobat/scripts/dcLana.js\')')
+    let inlineScript;
+    if (mobileWidget && request.device.isMobile) {
+      inlineScript = scripts
+        .replace('await import(\'./contentSecurityPolicy/csp.js\')', '{default:()=>{}}')
+        .replace('await import(\'./dcLana.js\')', 'await import(\'/acrobat/scripts/dcLana.js\')');
```
peller, line 132: DRY

edgeworkers/Acrobat_DC_web_prod/main.js
```diff
@@ -162,22 +172,28 @@ export async function responseProvider(request) {
...
   try {
-    const [
-      [responseStream, responseHeaders, dcCoreVersion],
-    ] = await Promise.all([
-      fetchFrictionlessPageAndInlineSnippet(),
-      fetchAndInlineScripts(),
+    const [responseStream, responseHeaders, dcCoreVersion, mobileWidget] = await fetchFrictionlessPageAndInlineSnippet();
+
+    await Promise.all([
+      fetchAndInlineScripts(mobileWidget),
```
peller, line 178: looks like this will hurt our performance. There are fetches in there that would otherwise be done in parallel with the first request

edgeworkers/Acrobat_DC_web_prod/main.js
```diff
@@ -162,22 +172,28 @@ export async function responseProvider(request) {
...
     const adobeid = isProd ? 'https://adobeid-na1.services.adobe.com' : 'https://adobeid-na1-stg1.services.adobe.com';
-    const headers = {
-      ...responseHeaders,
-      'Content-Security-Policy': csp,
-      Link: [
+
+    let headerLink;
+    if (mobileWidget && request.device.isMobile) {
+      headerLink = [
+        `<${adobeid}>;rel="preconnect"`,
```
peller, line 190: DRY

peller (changes requested):
need to make fetch requests in parallel where possible

edgeworkers/Acrobat_DC_web_prod/main.js
TsayAdobe: This update branches the condition `mobileWidget && request.device.isMobile` and otherwise. I feel that it's more readable with `mobileWidget && request.device.isMobile` and `!(mobileWidget && request.device.isMobile)`.

edgeworkers/Acrobat_DC_web_prod/main.js
TsayAdobe: We need to know if `mobile-widget` is true or false to decide what scripts to be inline and preloads. Therefore, they have to be separated.

peller: scripts.js is loaded regardless.

loading dc-conversion-widget.js conditionally, blocked by the first Promise, will hurt Frictionless performance on production where it is used. It might be better to load it, in parallel, and not use it where you don't need it.

peller:
Also, if we add a conditional to generate a different result for the same url, we can no longer cache it. That would be a huge performance loss.

edgeworkers/Acrobat_DC_web_prod/main.js
TsayAdobe: Fixed

edgeworkers/Acrobat_DC_web_prod/main.js
TsayAdobe: Fixed

TsayAdobe:
> Also, if we add a conditional to generate a different result for the same url, we can no longer cache it. That would be a huge performance loss.

onClientRequest adds device type to the cache key. Mobile, tablet, and desktop have different caches.

edgeworkers/Acrobat_DC_web_prod/main.js
peller: perhaps we could break fetchAndInlineScripts into two separate methods - fetchScript and inlineScripts. Then we could still fetch the frictionlesspage and fetchscripts in parallel, and then call inlineScripts(mobileWidget)

Blainegunn (approved):

peller (approved):

PR #802, opened 2024-09-13
MWPW-157166

Blainegunn:
Resolves : [MWPW-157166](https://jira.corp.adobe.com/browse/MWPW-157166)

We are removing the `EventListener` on our side for `change` and `drop` to allow the Unity Team to use it conflict free. 
In exchange they will dispatch a new event called `unity:track-analytics`. This event will also carry details on whether it's a **drop** or **change** event so we can keep our code functioning as it was before.
Thank you


saragajic11 (approved):

sanjayms01 (approved):

Ruchika4 (approved):

Ruchika4:
Thanks for the PR @Blainegunn as discussed over call unity will be sending an event so that DC can handle the analytics call for the events that have been commented. cc @sanjayms01

PR #819, opened 2024-09-23
implemented support for susi buttons

saragajic11:
Resolves: [MWPW-158072](https://jira.corp.adobe.com/browse/MWPW-158072)

Before: 
https://main--dc--adobecom.hlx.page/drafts/sara/susi

After: 
https://mwpw-158072-sign-in-link-support--dc--adobecom.hlx.page/drafts/sara/susi

Important: CORS is blocking the IMS:Ready event to be fired, so make sure you are using 'CORS unblock' or some other plugin to achieve the expected behaviour


acrobat/scripts/scripts.js
```diff
@@ -18,6 +18,10 @@ document.querySelectorAll('a').forEach((p, idx) => {
   }
 });
 
+document.querySelectorAll('a[href*="susi-"]').forEach((el) => {
+  el.style.display = 'none';
```
Blainegunn, line 22: We can set this as the default in `styles.css`.  
Instead of `display : none` let's do:
```
  pointer-events: none;
  cursor: default;
  ```
we can remove these styles in`handleImsSusi()`

File: acrobat/scripts/scripts.js [hunk 2]
```diff
@@ -460,6 +464,15 @@ const { ietf } = getLocale(locales);
 
   setConfig({ ...CONFIG, miloLibs });
 
+  window.addEventListener('IMS:Ready', async () => {
+    const susiSignUpElems = document.querySelectorAll('a[href*="susi-sign-up"]');
+    const susiSignInElems = document.querySelectorAll('a[href*="susi-sign-in"]');
+    if (susiSignUpElems.length > 0 || susiSignInElems.length > 0) {
```
Blainegunn, line 470: Let's make this more generic like you did above. Something like this:
```
const susiElems = document.querySelectorAll('a[href*="susi-"]')
```
then,  in `acrobat/scripts/susiAuthHandler.js` we can determine whether it's `-sign-up` or `-sign-in` and run the correct flow.

Blainegunn (commented):
The main reason for these changes is to reduce the number of `document.querySelectorAll` we call since they all hurt performance.

acrobat/scripts/scripts.js
saragajic11: resolved: https://github.com/adobecom/dc/pull/819/commits/98edddaaf1edd26392871a6e961d3625089cf653

acrobat/scripts/scripts.js
saragajic11: resolved: https://github.com/adobecom/dc/pull/819/commits/98edddaaf1edd26392871a6e961d3625089cf653

saragajic11:
> The main reason for these changes is to reduce the number of `document.querySelectorAll` we call since they all hurt performance.

thanks @Blainegunn !

Blainegunn (approved):

acrobat/scripts/alloy/susi-auth-handler.js
```diff
@@ -0,0 +1,17 @@
...
+      web: {
+        webInteraction: {
+          linkClicks: { value: 1 },
+          type: 'other',
+          name: `${ctaName}|marquee-${action}|overview|${device}`,
+        },
+      },
+    },
+  };
+  window?._satellite?.track('event', event);
```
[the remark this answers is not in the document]
joaquinrivero: Could you dix the eslint?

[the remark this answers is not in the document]
saragajic11: Fixed: https://github.com/adobecom/dc/pull/819/commits/0161ccb113ae95805ca1df39543de177614ae019
thanks @joaquinrivero !

joaquinrivero (approved):

PR #879, opened 2024-10-28
MWPW-161088 update

Blainegunn:
Resovles :  MWPW-161088


TsayAdobe (approved):
Actually you don't have to update the one in the `edgeworkers` folder. The deployment script will overwrite it with the one in the `acrobat` folder.

https://github.com/adobecom/dc/blob/stage/edgeworkers/scripts/build.js

PR #910, opened 2024-12-04
MWPW-161195 Reload page if going backward

TsayAdobe:
Reload the page if user is traversing the history.

Resolves: [MWPW-161195](https://jira.corp.adobe.com/browse/MWPW-161195)


joaquinrivero (approved):

tedAir (approved):
Verified on Chrome, FF, Safari. PASS.

PR #938, opened 2025-01-09
MWPW-163684 Set trial limit and show upsell

TsayAdobe:
It's the first part of the story. Social CTA will be added later after SUSI light block is done.

Resolve: [MWPW-163684](https://jira.corp.adobe.com/browse/MWPW-163684)

**Test URLs**
Before: https://www.stage.adobe.com/acrobat/online/test/compress-pdf.html
After: https://mwpw-163684--dc--adobecom.hlx.page/acrobat/online/test/compress-pdf?unitylibs=stage


acrobat/blocks/verb-widget/verb-widget.js
```diff
@@ -205,6 +236,14 @@ export default async function init(element) {
   // Analytics
   verbAnalytics('landing:shown', VERB);
 
+  if (LIMITS[VERB].trial) {
+    const count = parseInt(localStorage.getItem(`${VERB}_trial`), 10);
+    if (count >= LIMITS[VERB].trial) {
```
sanjayms01, line 241: If I'm understanding this correctly, this logic would show the upsell screen even on the 2nd attempt since it's using `>=`. Would `trial` need to be updated to 3 in the limits.js file?

acrobat/blocks/verb-widget/verb-widget.js
```diff
@@ -260,6 +300,12 @@ export default async function init(element) {
         setUser();
       },
       uploading: () => {
+        if (LIMITS[VERB].trial) {
+          const key = `${VERB}_trial`;
+          const stored = localStorage.getItem(key);
+          const count = parseInt(stored, 10);
```
sanjayms01, line 306:
```suggestion
          const count = stored !== null ? parseInt(stored, 10) : 0;
```

acrobat/blocks/verb-widget/verb-widget.js
TsayAdobe: 1st trial, count = NaN, after uploading, the local storage = 1. 
2nd trial, count = 1, after uploading,  the local storage = 2.
3rd attempt, count = 2,   `count >= LIMITS[VERB].trial`

acrobat/blocks/verb-widget/verb-widget.css
```diff
@@ -277,6 +278,66 @@
   text-decoration: none;
 }
 
+.verb-upsell {
+  color: rgb(34, 34, 34);
```
joaquinrivero, line 282: We could use a hex code here: #222222

acrobat/blocks/verb-widget/verb-widget.js
sanjayms01: Ah that makes sense, thanks for clarifying!

PR #955, opened 2025-01-16
MWPW-165646: Compress PDF - Supported file-type update

joaquinrivero:
Resolves: [MWPW-165646](https://jira.corp.adobe.com/browse/MWPW-165646)

Test URL:
- Before: https://stage--dc--adobecom.hlx.live/acrobat/online/compress-pdf
- After: https://mwpw-165646--dc--adobecom.hlx.live/acrobat/online/compress-pdf


TsayAdobe:
@joaquinrivero It's not a bug from the discussion. I think we want it to be as is.

joaquinrivero:
@TsayAdobe I added the list of file-type that is on compress-pdf now in dc-widget

TsayAdobe (approved):

PR #1003, opened 2025-02-14
MWPW-167953: Analytics variable update for totalSize and fileType

joaquinrivero:
Resolves: [MWPW-167953](https://jira.corp.adobe.com/browse/MWPW-167953)

Test URLs:
- Before: https://main--dc--adobecom.hlx.page/acrobat/online/compress-pdf
- After: https://mwpw-167953--dc--adobecom.hlx.page/acrobat/online/compress-pdf

### DIFF
--- a/acrobat/scripts/alloy/verb-widget.js
+++ b/acrobat/scripts/alloy/verb-widget.js
@@ -85,7 +85,7 @@ export default function init(eventName, verb, metaData, documentUnloading = true
               size: metaData?.size,
               count: metaData?.count,
               fileType: metaData?.type,
-              totalSize: metaData?.totalFileSize,
+              totalSize: metaData?.size,
             },
             source: {
               user_agent: navigator.userAgent,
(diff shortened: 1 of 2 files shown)


acrobat/scripts/alloy/verb-widget.js
```diff
@@ -82,8 +82,10 @@ export default function init(eventName, verb, metaData, documentUnloading = true
             },
             content: {
               type: metaData?.type,
-              size: metaData?.totalFileSize || metaData?.size,
+              size: metaData?.size,
               count: metaData?.count,
+              fileType: metaData?.type,
+              totalSize: metaData?.totalFileSize,
```
Ruchika4, line 88: totalSize and size should be sending the same data

Ruchika4 (changes requested):

acrobat/scripts/alloy/verb-widget.js
joaquinrivero: Ok, we will remove then the logic where the total size is calculated in our end

Ruchika4 (approved):

TsayAdobe (approved):

PR #1037, opened 2025-03-06
MWPW-169228 Set the default preview and live host

TsayAdobe:
Resolves: [MWPW-169228](https://jira.corp.adobe.com/browse/MWPW-169228)


mokimo (approved):

mokimo:
FYI this will need to be on the main branch, otherwise EDS will ignore those configs when I do the migration

tedAir (approved):

PR #1045, opened 2025-03-12
MWPW-169654 Optimizing Body Hiding for better Indexing and SEO

sharmrj:
As a Google bot, I need to access and read the full HTML of DC pages during the initial crawl, when JavaScript is disabled. Googlebot primarily crawls and indexes pages in two phases:

1.First pass (HTML-first indexing) – Googlebot attempts to index content without executing JavaScript, relying solely on the server-rendered HTML. If critical content is missing at this stage, it may not get indexed properly.

2.Second pass (JavaScript rendering) – Googlebot may revisit the page later with JavaScript enabled to process dynamically loaded content, but this can take time and is not guaranteed for all pages.

For optimal indexing, key content should be fully available in the server-rendered HTML during the first crawl to ensure visibility in search results.

Google bots currently see a blank page during indexing because the body is hidden using CSS. Since JavaScript is disabled when bots access the HTML, the content remains inaccessible.

This PR consumes a change in Milo introduced by https://github.com/adobecom/milo/pull/3804 to allow body hiding via javascript instead of css.

https://seo-body-unhide--dc--adobecom.hlx.live/acrobat/online/sign-pdf?milolibs=seo-body-unhide


Blainegunn (approved):

TsayAdobe:
DC has CSP. It needs the script hash in CSP.

Blainegunn:
> DC has CSP. It needs the script hash in CSP.

https://github.com/adobecom/dc/pull/1055

TsayAdobe (approved):

joaquinrivero (approved):

PR #1208, opened 2025-05-26
increase expiry timeout to 30 minutes while setting cookie

rohitsahu:
## Related Issue

Resolves: [MWPW-XXXXXX](https://jira.corp.adobe.com/browse/MWPW-XXXXXX)

## Test URLs

- https://main--dc--adobecom.aem.live/acrobat/online/compress-pdf
- https://mwpw-XXXXXX--dc--adobecom.aem.live/acrobat/online/compress-pdf

### DIFF
--- a/acrobat/blocks/verb-widget/verb-widget.js
+++ b/acrobat/blocks/verb-widget/verb-widget.js
@@ -1112,7 +1112,7 @@ export default async function init(element) {
   });
 
   element.addEventListener('unity:track-analytics', (e) => {
-    const cookieExp = new Date(Date.now() + 90 * 1000).toUTCString();
+    const cookieExp = new Date(Date.now() + 30 * 60 * 1000).toUTCString();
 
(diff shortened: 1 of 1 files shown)


Blainegunn (changes requested):
We need `domain=.adobe.com;` 
Why do you need it removed?

rohitsahu:
> We need `domain=.adobe.com;` Why do you need it removed?

@Blainegunn Venkata needs to try out an extensive upload scenarios with varied file sizes for backend analysis around upload time to improve the upload success rate. A private branch has been created which points to backend dev environment for this activity. 
Private branches only work with the hlx pages. The uploading and uploaded time is not getting set in the cookie due to the domain check. 
Given that we can't remove the`domain=.adobe.com` check, does it work if we change thisnon stage for couple of days and provide @sigadamvenkata the window to perform the activity before we revert it?

PR #1242, opened 2025-06-19
MWPW-174349 (Refactor)

Blainegunn:
## Related Issue

Resolves: [MWPW-174349](https://jira.corp.adobe.com/browse/MWPW-174349)

## Test URLs

- https://main--dc--adobecom.aem.live/acrobat/online/compress-pdf
- https://mwpw-174349--dc--adobecom.aem.live/acrobat/online/compress-pdf


joaquinrivero (commented):
What about check if there are not '' in case someone remove the content but not the row

Blainegunn:
> What about check if there are not '' in case someone remove the content but not the row

@joaquinrivero it will come through blank. They could be a valid use case, maybe?
I don't think we've ever enforced this. I'm fine leaving as is and letting GWP be accountable for their content.

PR #1297, opened 2025-09-29
DC MILO Release 0.2538

Ruchika4:
PR:

- https://github.com/adobecom/dc/pull/1296


ivanvatadobe:
noting that this has been verified in lower envs and stage for the test page it is being enabled for.

